PR #5, opened 2021-04-29
Add practical node tests (lodash...)

evrardjp:


evrardjp:
This seems broken on my system, the image doesn't seem to have git.

evrardjp:
Retriggered testing, as the containers have changed now

evrardjp:
Rebased with latest poetry merges.

dcermak:
@evrardjp Would you take a look at what I've done to your node tests (unfortunately I cannot request a review from you, as you submitted the PR)? The CI *should* be passing now :crossed_fingers:.

test_node.py
```diff
@@ -1,19 +1,133 @@
...
+                build_command="npm ci && npm test && npm run lint",
+            ),
+            NpmPackageTest(
+                repository_url="https://github.com/visionmedia/debug.git",
+                build_command="npm install && npm run lint && npm run test:node",
+            ),
+            NpmPackageTest(
+                repository_url="https://github.com/expressjs/express.git",
+                build_command="""npm config set shrinkwrap false &&
+            npm rm --silent --save-dev connect-redis &&
```
evrardjp, line 80: Ok the only thing I find to say is that those lines are ugly, and would be better if they were all indented 4 spaces below build_command.

test_node.py
```diff
@@ -1,19 +1,133 @@
...
+
+    @property
+    def test_command(self) -> str:
+        return f"""git clone {self.repository_url} &&
+cd {self.repo_name} &&
+{self.build_command}
+"""
+
+    def __str__(self) -> str:
+        return self.repo_name
```
evrardjp, line 36: that will be a long name, I would extract just the last word of the the url instead (basename).

test_node.py
```diff
@@ -1,19 +1,133 @@
...
+                    repository_url="https://github.com/facebook/prop-types",
+                ),
+                pytest.mark.skip(reason="Broken for some reason"),
+            ),
+            (
+                NpmPackageTest(
+                    repository_url="https://github.com/jprichardson/node-fs-extra",
+                    build_command="npm install && npm run unit",
+                ),
+                pytest.mark.xfail(
```
evrardjp, line 122: nit: I can't shake why I would do it, but I would pass the marks as kwargs of NpmPackageTest, and just do a single list comprehension for all the tests `[ pkg.to_pytest_param(marks=self.marks) for pkg in (NpmPackageTest(build_command=, repository_url=, marks=` marks being only set  on L103+ . WDYT? Will it be more readable for non pythonistas?

test_node.py
dcermak: No, `repo_name` is e.g. `react` or `lodash`.

test_node.py
dcermak: Yeah, originally I wanted to treat `NpmPackageTest` as a pure data container, but now that it also contains some "business logic", I think adding the mark there makes sense as well.

test_node.py
dcermak: Now better?

dcermak:
For some reason the react.js tests are failing on github actions but not locally for me (at least not when I run them serially). I guess that the react tests are too fat and fail if you overload the host. So either we don't run them in parallel, or we just drop them from the test matrix. I'm actually more inclined to dropping react, as it will make the test matrix really huge.

test_node.py
evrardjp: I see now it is using basename, good.

evrardjp:
Well, github actions isn't using `tox --parallel`, it's only doing `tox -e node`. However, `tox -e node` is running the `test_node.py` tests in parallel, which might be the cause of conflict when multiple of those scripts run on the same machine. Maybe we can change the scope of the fixture to move to a function one, but it will be by far less efficient.

I am fine with dropping react, but we'll have to think about those cases in the future.

dcermak:
I've opted to skip the react tests for now until we figure out how we can run certain tests serially and others in parallel.

PR #7, opened 2021-04-30
[WIP] Add pack support

evrardjp:
This will add testing of the `pack` command, starting with what
was recommended on paketo website (using buildpacks.io).

-----
[View rendered README.md](https://github.com/evrardjp/m8a-tests/blob/test-pack/README.md)

### DIFF
--- a/test_openjdk.py
+++ b/test_openjdk.py
@@ -15,6 +15,7 @@ def test_maven_present(container):
     assert container.connection.run_expect([0], "mvn --version")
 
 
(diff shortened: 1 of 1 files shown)


evrardjp:
That needs rethinking, with the latest changes in our repo. We could run pack inside container, but that sounds a hassle.

Instead I would think to rely on pack from the host (we'll need a separate github action job for that).

dcermak:
stale PR

PR #11, opened 2021-05-04
Use poetry

dcermak:


tox.ini
```diff
@@ -2,19 +2,19 @@
...
 commands =
-    m8a-fetch-language-containers {envname}
+    poetry run m8a-fetch-language-containers {envname}
     # -n auto tells pytest to use as many processors as detected
     # to run all the tests from the module test_{envname} in parallel
-    pytest -n auto test_{envname}.py []
+    poetry run pytest -n auto test_{envname}.py []
 
 [testenv:fetch-all]
 commands =
```
evrardjp, line 20: Can you update the poetry commands here too?

evrardjp (changes requested):
I am okay with poetry. However, that means that we need to install tox+poetry in the host, instead of tox+setuptools. In other words, it means the github actions need updating to install poetry.

tox.ini
dcermak: Like this?

evrardjp: Yup!

evrardjp (approved):
It's a brave new world!

evrardjp (changes requested):
Oh well, I forgot that we need to add poetry into gh actions.

evrardjp (approved):
I guess ` +1,299 −33 ` is what we have to pay for reproducibility.

Ok let's try this.

PR #16, opened 2021-05-04
Fix container names

dcermak:
The python and node containers were using the wrong repositories.


evrardjp:
We could have done 2 PRs for this, but I think I will survive the two different scope in a single PR.

Let's wait for those containers to be reachable to retrigger tests and merge this.

evrardjp (approved):
LGTM, now the obs waiting game starts.

PR #20, opened 2021-05-05
Restructure test data

dcermak:
This changes how the `containers` dictionary is created in the `data` module. While this is quite a fat change with no immediately visible benefit, it will become handy later on, as we can query each container for its project, version, repository, etc.pp. I am currently working on a rabbitmq listener that will automatically run the CI if OBS publishes a container to the the repositories (this would be doable with the previous dictionary, but super ugly imho).

### DIFF
--- a/matryoshka_tester/data.py
+++ b/matryoshka_tester/data.py
@@ -1,28 +1,128 @@
-containers = dict()
-containers["go"] = {
+from dataclasses import dataclass
+from typing import Dict, List, Optional
+
+
+go_containers = {
     "1.16": "registry.opensuse.org/devel/bci/images/bci/golang-devel:16",
     "1.15": "registry.opensuse.org/devel/bci/images/bci/golang-devel:15",
-    "1.14": "registry.opensuse.org/devel/bci/images/bci/golang-devel:14"
+    "1.14": "registry.opensuse.org/devel/bci/images/bci/golang-devel:14",
 }
-containers["node"] = {
+node_containers = {
     "15": "registry.opensuse.org/home/fcrozat/matryoshka/containers_node15/node15:latest",
     "14": "registry.opensuse.org/home/fcrozat/matryoshka/containers_node14/node14:latest",
     "12": "registry.opensuse.org/home/fcrozat/matryoshka/containers_node12/node12:latest",
 }
-containers["openjdk-devel"] = {
+openjdk_devel_containers = {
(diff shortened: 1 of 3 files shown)


matryoshka_tester/data.py
```diff
@@ -1,27 +1,108 @@
...
-    "16": "registry.opensuse.org/devel/bci/images/bci/openjdk:16",
-    "15": "registry.opensuse.org/devel/bci/images/bci/openjdk:15",
-    "14": "registry.opensuse.org/devel/bci/images/bci/openjdk:14",
-    "11": "registry.opensuse.org/devel/bci/images/bci/openjdk:11",
-}
-containers["python"] = {
-    "3.9": "registry.opensuse.org/home/fcrozat/matryoshka/containers_python39/python39:latest",
-    "3.8": "registry.opensuse.org/home/fcrozat/matryoshka/containers_python38/python38:latest",
-    "3.6": "registry.opensuse.org/home/fcrozat/matryoshka/containers_python36/python36:latest",
+from dataclasses import dataclass
```
evrardjp, line 1: What I liked in this file, was the very simple way for a non-experienced user to iterate here, add its own container, change it. I am not sure if this is still as readable (it is okay for me, but I admit it's not as clear).

If I were to rebuild this, I would probably make sure that we have data in a certain form, and (un)marshall it into the proper python objects (this would be the way to build the `containers` dict [maybe we could move this to a list?]). The objects could be `Container`s, for which we could have different implementations, based on the input data. e.g. we could inherit the container for specific language containers (`GoContainer`?). I feel like this would make more sense and be more readable.  WDYT? 

In a certain way, I wanted to separate the _data_ from the code.

Side note: ContainerConfig would be removed too if we had this, and make the whole purpose of the ContainerConfig (type validation) far more clear, IMO.

evrardjp (changes requested):
Type validation gets my <3. 

I believe we should move a few of the data structures into something else than `data`.py. I feel like there is something that doesn't sound right.

However, the code itself is good.

I wonder if you could tell me more about the rabbitmq consumer, maybe it will help me understand where you go from here.

matryoshka_tester/data.py
dcermak: Yeah, I must confess that the part of my code that I really hate is how it makes the registry URL addition super confusing. Unmarshalling it sounds way better, so I'll give that a try.

evrardjp:
I am closing this, as we have an "alt" PR that's also working and I find clearer. Sorry! :)

PR #24, opened 2021-05-06
Add venv for commands

evrardjp:
Without this, one needs to enter the venv to run specific command
that are installed through setuptools entrypoints.

This is mildly annoying.

This fixes it by introducing a 'venv' environment, whose purpose
is only to accept commands.

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -23,3 +23,7 @@ commands =
 [testenv:list-all]
 commands =
     m8s-list-containers
+
+[testenv:venv]
+passenv = *
+commands = {posargs}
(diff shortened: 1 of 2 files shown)


tox.ini
```diff
@@ -23,3 +23,6 @@ commands =
 [testenv:list-all]
 commands =
     m8s-list-containers
+
+[testenv:venv]
+commands = {posargs}
```
dcermak, line 28: Without this some "fatter" applications will not start, as stuff like `TERM` or `LANG` are not set.
```suggestion
passenv = *
commands = {posargs}
```

dcermak (changes requested):
I need the passenv entry to launch emacs via tox. I cannot live without emacs 😜

tox.ini
evrardjp: Updated the code and commit message.

dcermak (approved):

PR #25, opened 2021-05-06
[CI] Change the location of docker's data-dir to cache images

dcermak:
This PR adds a cache of docker's image layer directory, hopefully speeding up subsequent image pulls.


.github/workflows/ci.yaml
```diff
@@ -33,11 +33,21 @@ jobs:
...
+      - name: change the location of the docker data dir to cache images
+        run: |
+          mkdir -p ./data-dir
+          cat /etc/docker/daemon.json | jq --arg PWD $(pwd) '. + { "data-root": ($PWD + "/data-dir") }' | sudo tee /etc/docker/daemon.json
+          sudo systemctl restart docker
+      - name: cache docker images
+        uses: actions/cache@v2
+        with:
+          path: ./data-dir/
+          key: ${{ runner.os }}-docker-images
```
evrardjp, line 47: If cache is shared between two jobs, I think it makes sense to fetch all images in the job "gentestmatrix". 
Then, the parallel jobs can use the cache already set in place.

dcermak:
The github action is refusing to cache the images :(

PR #29, opened 2021-05-07
[CI] Add trivy vulnerability scan

dcermak:
Let's have another security scanner :wink:


dcermak:
Note that the CI failure is due to the security scanning stuff hitting github's API rate limit. Not sure how to handle this properly…

PR #31, opened 2021-05-10
Serial tests

dcermak:
This PR adds the `serial` mark to pytest so that certain tests can be executed in serial.

### DIFF
--- a/test_node.py
+++ b/test_node.py
@@ -92,12 +92,6 @@ def test_npm(container):
                 repository_url="https://github.com/caolan/async",
             ),
             NpmPackageTest(
-                build_command="yarn --frozen-lockfile && yarn test",
-                repository_url="https://github.com/facebook/react.git",
-                repository_tag="v17.0.2",
-                marks=pytest.mark.serial,
-            ),
-            NpmPackageTest(
                 build_command="""yarn install &&
                     npm install --no-save "react@~16" "react-dom@~16" &&
                     yarn run pretest &&


tox.ini
```diff
@@ -14,7 +14,8 @@ commands =
     m8a-fetch-language-containers {envname}
     # -n auto tells pytest to use as many processors as detected
     # to run all the tests from the module test_{envname} in parallel
-    pytest -n auto test_{envname}.py --junitxml={toxinidir}/junit.xml []
+    pytest -n auto -m "not serial" test_{envname}.py --junitxml={toxinidir}/junit_parallel.xml []
+    pytest -n0 -m "serial" test_{envname}.py --junitxml={toxinidir}/junit_serial.xml []
```
evrardjp: You can remove -n0 . I think it's clearer, and that way the pytest-xdist isn't involved at all, iirc.

evrardjp (approved):
I like this, it's clear.

tox.ini
evrardjp: Meh, let's ignore this ^

evrardjp:
Waiting for node to return green beffore pressign the big green button.

evrardjp (changes requested):
I would prefer if we update the commit message for the add pytest serial mark.

evrardjp (dismissed):
LGTM

dcermak:
The react.js tests are consistently failing in github actions, I'd drop them because they don't add any value imho. WDYT @evrardjp?

dcermak:
Jean-Philippe Evrard ***@***.***> writes:

> Yeah let's remove them. We have enough examples. 

I already removed them.

PR #33, opened 2021-05-10
Cleanup the test_node file

evrardjp:
Without this, documentation seems intertwined in the test_node
file, where the documentation should be complete in the README
instead.

This should fix it.

At the same time, I am cleaning up so that black is happy.

### DIFF
--- a/README.md
+++ b/README.md
@@ -15,13 +15,36 @@ easy.
 
 You can use the convenience tools from conftest:
 
-* If you are using the "container" fixture, your test will auto generate the right tests for _all_ the versions of your language. This is auto loaded, and doesn't need anything from your side except using the keyword "container"
-* If you want to _skip_ some of those tests, use the decorator named `restrict_to_version`. This decorator accepts a list of strings matching the versions from `data.py`. To use it, add first `from conftest import restrict_to_version` and then wrap your code like this (assuming openjdk here):
+* If you are using the "container" fixture, your test will auto generate the right tests for _all_ the versions of your language. This is auto loaded, and doesn't need anything from your side except using the keyword "container". See below for more details.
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -15,14 +15,38 @@ easy.
...
+The container fixture automatically finds the testfile filename, uses it to guess the language of the container under test,
+and starts all the necessary containers. See also conftest.py
+
+### The restrict_to_version decorator
+
+This decorator accepts a list of strings matching the versions from `data.py`.
+
+To use it:
+
+1. add `from conftest import restrict_to_version` in your imports.
```
dcermak, line 39:
```suggestion
1. add `from conftest import restrict_to_version` to your imports.
```

README.md
```diff
@@ -15,14 +15,38 @@ easy.
...
+2. wrap your code like this (assuming openjdk here):
+
 ```Python
 @restrict_to_version(['11'])
 def mytest(container):
     pass
 ```
 
+This example above means that, from all the containers for openjdk, the `mytest` test will only run for openjdk '11' container.
```
dcermak, line 48:
```suggestion
In the example above only the `openjdk:11` container will be used to run the test function `mytest`.
```

README.md
```diff
@@ -15,14 +15,38 @@ easy.
...
+and starts all the necessary containers. See also conftest.py
+
+### The restrict_to_version decorator
+
+This decorator accepts a list of strings matching the versions from `data.py`.
+
+To use it:
+
+1. add `from conftest import restrict_to_version` in your imports.
+2. wrap your code like this (assuming openjdk here):
```
dcermak, line 40:
```suggestion
2. wrap your test function as follows (assuming openjdk here):
```

dcermak (dismissed):
I've added a few nagging comments to the documentation changes.

evrardjp:
Will fix those now.

PR #35, opened 2021-05-11
Format and misc fixes

dcermak:
This is a collection of miscellaneous fixes that were not too important to block a review of a previous PR, but have accumulated by now.
Also, I have added black to tox and ran it on the source code, so https://github.com/SUSE/m8a-tests/commit/93e3d139ac130af1f0b9bd4a808e6201f4ac5848 is a big diff without functional changes.


evrardjp (approved):
<3

PR #39, opened 2021-05-11
Restructure test data (alt)

evrardjp:
This is using a `containers.json` data, on which we unserialize data from, and build our containers list.

PR #43, opened 2021-05-12
Rabbitmq trigger

dcermak:
This PR is based on #39: it adds a new entry point that listens to rabbit.opensuse.org and launches the CI pipeline every time the repositories get published.

### DIFF
--- a/matryoshka_tester/trigger_test.py
+++ b/matryoshka_tester/trigger_test.py
@@ -1,0 +1,175 @@
+from dataclasses import dataclass, field
+import json
+from logging import getLogger
+from os import environ
+from typing import (
+    Dict,
+    Literal,
+    List,
+    Optional,
+    NoReturn,
+    Tuple,
+    TypedDict,
+    Union,
+)
+
+import pika
+import requests
+
+from matryoshka_tester.parse_data import containers, Container
+
+
+LOGGER = getLogger(__name__)
+GITHUB_API_BASEURL = "https://api.github.com"
+
+
+class PublishBodyBase(TypedDict):
+    #: name of the project which' publishing state changed
+    project: str
+    #: repository which' publishing state changed
+    repo: str
+
+
+class PacktrackBody(PublishBodyBase):
+    #: some id identifying something
+    payload: str
+
+
+class PublishStateBody(PublishBodyBase):
+    #: the new state of the project + repository
+    state: Literal[
+        "unknown",
+        "broken",
+        "scheduling",
+        "blocked",
+        "building",
+        "finished",
+        "publishing",
+        "published",
+        "unpublished",
+    ]
+
+
+class PublishedBody(PublishBodyBase):
+    #: some internal (=undocumented) build ID
+    buildid: str
+
+
+def trigger_workflow(
+    event_type: str = "containers_published",
+    owner: str = "SUSE",
+    repo: str = "m8a-tests",
+):
+    headers = {
+        "Accept": "application/vnd.github.v3+json",
+        "Authorization": f"token {environ['GITHUB_TOKEN']}",
+    }
+
+    try:
+        resp = requests.post(
+            f"{GITHUB_API_BASEURL}/repos/{owner}/{repo}/dispatches",
+            json={"event_type": event_type},
+            headers=headers,
+        )
+
+        if resp.status_code != 204:
+            LOGGER.error("triggering the workflow failed with %s", resp.status)
+    except Exception as exc:
+        LOGGER.error("sending the workflow trigger failed with %s", exc)
+
+
+@dataclass
+class RabbitMQConnection:
+    #: url to the message bus, defaults to openSUSE's
+    url: str = "amqps://opensuse:opensuse@rabbit.opensuse.org"
+    #: the rabbitMQ prefix used by OBS, see
+    #: https://openbuildservice.org/help/manuals/obs-admin-guide/obs.cha.administration.html#_message_bus
+    message_topic_prefix: str = "opensuse.obs"
+
+    #: list of routing keys corresponding to the repository publishing events
+    repository_routing_keys: List[str] = field(default_factory=list)
+
+    #: Dictionary of project names (as keys) mapping to a list of repositories
+    #: whose publishing event triggers a CI run on github
+    watched_projects: Dict[str, List[str]] = field(default_factory=dict)
+
+    container_list: Optional[List[Container]] = None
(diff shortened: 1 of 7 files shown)


matryoshka_tester/trigger_test.py
```diff
@@ -0,0 +1,119 @@
...
+        "Authorization": f"token {environ['GITHUB_TOKEN']}",
+    }
+
+    resp = requests.post(
+        f"{GITHUB_API_BASEURL}/repos/{owner}/{repo}/dispatches",
+        json={"event_type": event_type},
+        headers=headers,
+    )
+
+    if resp.status_code != 204:
```
evrardjp, line 94: Is that normal that you're not raising an error here?
I feel like throwing an exception here would make the reason of handling the exception more explicit to where the error bubbles. e.g. here in the 'callback' method, we decide or not to return an empty result, and error, or crash if really necessary (instead of just throwing an error message).

matryoshka_tester/trigger_test.py
```diff
@@ -0,0 +1,119 @@
...
+        "unpublished",
+    ]
+
+
+class PublishedBody(PublishBodyBase):
+    #: some internal (=undocumented) build ID
+    buildid: str
+
+
+connection = pika.BlockingConnection(
```
evrardjp, line 43: I don't know why, but I would move all of this into an init function, which takes certain parameters, so that it's easily unit-testable.

Next to this, if we have another way to trigger_tests in the future that doesn't depend on mqtt, we won't do those unnecessary initializations.

matryoshka_tester/trigger_test.py
```diff
@@ -0,0 +1,119 @@
...
+
+PREFIX = "opensuse.obs"
+REPO_ROUTING_KEYS = [
+    # f"{PREFIX}.repo.packtrack",
+    f"{PREFIX}.repo.publish_state",
+    f"{PREFIX}.repo.published",
+]
+
+
+watched_projects: Dict[str, List[str]] = {}
```
evrardjp, line 66: same here, I would just put this into a function for unit testing.
This means the watched projects need to be passed around, in the 'main' ('trigger_from_mqtt') function.

Reducing the global variables will things also explicit for callback.

matryoshka_tester/trigger_test.py
```diff
@@ -0,0 +1,119 @@
...
+            return
+
+        if payload is None or payload["project"] not in watched_projects:
+            return
+
+        if payload["repo"] in watched_projects[payload["project"]]:
+            trigger_workflow("containers_published")
+
+
+GITHUB_API_BASEURL = "https://api.github.com"
```
evrardjp, line 112: nit: I would move this a bit higher, as we know it's a static var. But it's not necessary to do a commit just for that though :)

tox.ini
```diff
@@ -11,10 +11,13 @@ deps =
     pytest-custom_exit_code
     prettytable
     black
+    pika
```
evrardjp, line 14: nit:
maybe at some point we'll have to build a `requirements.txt`, and or a proper series of requirements for setuptools, including different optional modules.

In this case, pika/requests might probably go into the trigger-tests environment, but I think it can be done later.

evrardjp (commented):
I am not sure if it's good to _request changes_ on something that works, so here are my comments.
Feel free to adapt the unit testing in this PR or in a follow up one.

tox.ini
dcermak: I'm a bit torn on this one: on one hand it would be nice only include pika & requests where required. On the other hand, it means that if they are missing from the default venv, then any tooling launched via `tox -e vevn -- $cmd` is not aware of them and these modules are not available for unit testing. I found the later to outweigh the former and thus opted for it.

evrardjp: It was a nit, so it's no big deal.

I think it would be better if I show you what I mean with a follow-up PR, and we discuss it there.

test_unit.py
```diff
@@ -36,3 +35,45 @@ def test_host_fips_disabled(tmp_path):
...
+                type="node",
+                repo="home/fcrozat/matryoshka/containers_node15",
+                image="python15",
+                tag="latest",
+                version="15",
+            ),
+        ]
+    )
+    print(con.watched_projects)
+    assert con.watched_projects["devel:bci"] == ["images"]
```
evrardjp, line 74: That will probably break soon, but I expect it's a good thing?

test_unit.py
```diff
@@ -36,3 +35,45 @@ def test_host_fips_disabled(tmp_path):
     fipsfile = tmp_path / "fips"
     fipsfile.write_text("")
     assert not host_fips_enabled(f"{fipsfile}")
+
+
+def test_routing_keys():
+    con = RabbitMQConnection(message_topic_prefix="foo.bar")
```
evrardjp, line 41: isn't there a risk of failing unit tests with there are problems to reach opensuse? 
I am wondering if we should mock here, instead.

evrardjp (commented):
WDYT?

evrardjp:
The code looks good and can be merged.
For the unit tests, we can discuss it in a follow up commit if we don't reach consensus.

@dcermak Can you rebase this branch, and comment on all my comments above?

matryoshka_tester/trigger_test.py
dcermak: Yes, because if I throw an exception here, then it will kill the whole script. I have now switched to logging that case instead.

matryoshka_tester/trigger_test.py
dcermak: I have moved this into the function `RabbitMQConnection.connect_channel_to_bus`

matryoshka_tester/trigger_test.py
dcermak: Done, this is now in the class attribute `RabbitMQConnection.watched_projects`

tox.ini
dcermak: sure, sounds good to me.

test_unit.py
dcermak: No, this unit test does not connect to the bus at all.

test_unit.py
dcermak: No, it won't break, because this is calculated from the list in line 49.

dcermak:
> @dcermak Can you rebase this branch, and reply on all my comments above?

I have rebased the branch

evrardjp (approved):

PR #44, opened 2021-05-12
Allow to run tooling without podman

evrardjp:
Without this patch, we try to instantiate PODMAN_RUNTIME at all
times.

This is a problem, as some user might not have PODMAN, and the
program will crash with a Runtime Error.

This fixes it by only instantiating a podman runtime if it's
worth it.

### DIFF
--- a/matryoshka_tester/helpers.py
+++ b/matryoshka_tester/helpers.py
@@ -92,10 +92,6 @@ class DockerRuntime(OciRuntimeBase):
         return last_line.split()[-1]
 
 
-DOCKER_RUNTIME = DockerRuntime()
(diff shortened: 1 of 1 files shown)


dcermak (changes requested):
Please mention that this reverts f3ed2a6686b022fab286b95a0204931bad9f8534

dcermak (approved):
Thanks! And sorry for messing this one up…

evrardjp:
You shouldn't worry about breaking things :)
Else it will prevent you from iterating on things. Breaking things is a way to success! 

The only thing I care is about whether things are fixed when they are broken, and we've done that now :)

PR #45, opened 2021-05-12
Remove async from test_node

dcermak:
The async package ocasionaly fails to test, but often enough to make the CI flaky.


evrardjp (approved):
I don't see a problem to remove this for now.

PR #48, opened 2021-05-12
[CI] Only run podman tests

evrardjp:
The docker tests are the same, so it's not really worth running.
If a contributor runs docker without podman, that will be enough.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -47,7 +47,6 @@ jobs:
       matrix:
         toxenv: ${{fromJson(needs.gentestmatrix.outputs.matrix)}}
         container_runtime:
-          - DOCKER
           - PODMAN
     steps:
     - name: checkout source code


dcermak:
I'm not sure if we really want to go down this route? Docker is still the most popular container runtime, so maybe we shouldn't just drop it from the CI.

evrardjp:
Yeah but the CI is using a docker emulation from podman, right? It's not like it contains the docker community code?

evrardjp:
Maybe I should remove podman, and install docker on that test?

evrardjp:
Closing this temporarily, I will need to think about it : )

dcermak:
Jean-Philippe Evrard ***@***.***> writes:

> Yeah but the CI is using a docker emulation from podman, right? It's not like it contains the docker community code?

No, I don't think that it uses podman's docker emulation. That one is
only used if you install podman-docker and remove docker
itself (neither is the case on github action runners).

evrardjp:
Fair.

PR #49, opened 2021-05-14
Mark the visionmedia/debug build as failure

evrardjp:
Without this, node tests break.

This is a problem, as we are making those mandatory.

This fixes it by marking the test as _expected failure_ instead
of skipped, so that we can track when it's working again, and
re-enable it.

### DIFF
--- a/test_node.py
+++ b/test_node.py
@@ -39,6 +39,9 @@ def test_npm(container):
                     npm run lint &&
                     npm run test:node
                     """,
+                marks=pytest.mark.xfail(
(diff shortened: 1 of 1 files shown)


dcermak:
Will xfail cause the test suite to fail if the test stops failing?

evrardjp:
Rebased.

evrardjp:
> 
> 
> Will xfail cause the test suite to fail if the test stops failing?

Yes, on purpose! That will allow us to track when things are working again.

dcermak (approved):

PR #52, opened 2021-05-14
Not a known digest is in stderr

evrardjp:
Without this, the assertion fails, because the error message
appears on stderr, not stdout.

This fixes the FIPS tests.


dcermak (approved):
Works in my FIPE enabled VM :tada:

PR #61, opened 2021-05-17
Signature verification

dcermak:
This PR extends the current CI to use podman's signature verification feature.


dcermak:
The CI is failing at the moment, because not all containers are build from devel:BCI.

PR #62, opened 2021-05-17
Building Fleet

mvarlese:


dcermak (changes requested):
The CI fails with: `/bin/sh: curl: command not found`

mvarlese:
> The CI fails with: `/bin/sh: curl: command not found`

Yes, I'm aware. I'm adding the missing packages onto the repository for the BCI... The PR was issued nonetheless as per chat with @evrardjp

evrardjp:
Maybe the command can be extended by a `zypper install <all the packages you need>`. That's effectively what would be done in a dockerfile or by using another base container. This also proves how efficient our go container is.

mvarlese:
> Maybe the command can be extended by a `zypper install <all the packages you need>`. That's effectively what would be done in a dockerfile or by using another base container. This also proves how efficient our go container is.

Yes, that's idea @evrardjp 
However, before I submit the new PR I am checking locally that the new repo contains the packages required by zypper in the first place. The current status of the repo is *broken* so I cannot validate my local revision yet.
That said, please, stay tuned. :)

mvarlese:
Another one coming

PR #66, opened 2021-05-18
Fix and simplify CI

evrardjp:
Without this, the container-scan is incorrectly having the version
instead of the URL to test.

This is a problem, as it will prevent those tests to run correctly.

This fixes it by simplifying the stringified_matrix generation,
adapting to the right filed.

Next to this, the ci is using tox, but we have established a list
of what we test inside the CI that's not matching tox, for
simplicity.

Because we are having more and more tests, I simplified so that
the CI is matching tox envlist. Extra environments won't be
tested, as it is done currently.


dcermak (approved):
Looks good to me, thanks!

PR #67, opened 2021-05-18
Add test to build fleet using dapper

dcermak:


evrardjp (approved):

test_go.py
```diff
@@ -19,5 +21,18 @@ def test_go_version(container):
...
+    [
+        GitRepositoryBuild(
+            repository_url="https://github.com/mvarlese/fleet",
+        ).to_pytest_param()
+    ],
+    indirect=["host_git_clone"],
+)
+def test_fleet(host, host_git_clone, dapper):
+    dest, _ = host_git_clone
+    host.run_expect([0], f"cd {os.path.join(dest, 'fleet')} && {dapper}")
```
evrardjp, line 38: shouldn't this be dapper make ?

evrardjp (changes requested):
I think this is not doing the right thing, except if I am misunderstanding.

test_go.py
mvarlese: In my local tests, I simply run "make" from the top-level folder in the fleet project indeed.

dcermak: It should be either `dapper build` or `dapper ci`.

evrardjp (approved):
Ok tracked it on the fleet repo, makes sense to me.
You have to follow the path towards the Dockerfile of fleet to understand this. It surely isn't obvious at first glance, and IMO should not be necessary. Anyway, this makes this tool tested, and should make some folks happy, I guess.

PR #72, opened 2021-05-20
[WIP] add a simple size check

dcermak:
This currently only checks the base container

### DIFF
--- a/test_size.py
+++ b/test_size.py
@@ -1,0 +1,67 @@
+from typing import Union
+
+import pytest
+
+
+@pytest.mark.parametrize(
+    "image_name,size_mb",
+    [
+        ("bci/base", 100),
+        pytest.param(
+            "bci/golang:1.16",
+            1000,
+            marks=pytest.mark.xfail(reason="The golang container is too fat"),
+        ),
+        pytest.param(
+            "bci/golang:1.15",
+            1000,
+            marks=pytest.mark.xfail(reason="The golang container is too fat"),
+        ),
+        pytest.param(
+            "bci/golang:1.14",
+            1000,
+        ),
+        pytest.param(
+            "bci/openjdk-devel",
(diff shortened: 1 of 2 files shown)


tox.ini
```diff
@@ -1,5 +1,5 @@
 [tox]
-envlist = unit, base, python, node, go, openjdk, openjdk-devel, multistage, fetch-all, list-all
+envlist = unit, base, python, node, go, openjdk, openjdk-devel, multistage, size, fetch-all, list-all
```
evrardjp, line 2: we'll need to change tox. Either its own env to not fetch language containers matching the name 'size'. Alternatively, we can probably change the default testenv (so that it doesn't run commands). If we rename the language container tests, we might have something simpler.

test_size.py
```diff
@@ -0,0 +1,29 @@
+from matryoshka_tester.parse_data import containers
+
+
+def test_size(host, container_runtime):
+    all_images = filter(
+        None,
+        host.run_expect(
+            [0],
+            f"{container_runtime.runner_binary} images "
+            + '--format "{{ .Repository }}:{{ .Tag }} {{ .Size }}"',
```
evrardjp, line 10: there is a filter argument to the docker images command that will make the get_container_by_image_name disappear.

evrardjp (changes requested):

test_size.py
dcermak: Yes, that's a good idea. I've changed it.

tox.ini
dcermak: Yeah, we need to change something here, because at the moment the tests fail as the images are not present. Would it be feasible to run fetch-all beforehands? Eventually we'd want to test all containers anyway, right?

dcermak:
Superseded by #74

PR #74, opened 2021-05-21
Add a way to test decompressed size of containers

evrardjp:
Without this, we don't ensure any size of containers.
This is a first step, tackling the decompressed size,
until we have other tests to tackle the compressed
size present in the registry.


dcermak (approved):
Much better than my version!

PR #76, opened 2021-05-26
Use NodeJS container images published by devel:BCI

mvarlese:


test_go.py
```diff
@@ -23,13 +23,13 @@ def test_go_version(container):
     "container_git_clone",
     [
         GitRepositoryBuild(
-            repository_url="https://github.com/weaveworks/kured.git",
```
dcermak, line 26: Why are you changing the golang tests?

mvarlese: hmmm.... i guess my repo is out of sync somehow :(

PR #114, opened 2022-05-05
Enable tests for openjdk-17 container

jlausuch:


dcermak (changes requested):
These modifications wouldn't catch a wrongly set Java version in the environment variable `JAVA_VERSION` if the installed java is also wrong.

jlausuch:
> These modifications wouldn't catch a wrongly set Java version in the environment variable `JAVA_VERSION` if the installed java is also wrong.

Right. Then, we also need to correct python tests, as they are doing this exact same thing.

The problem I had while running the tests with
```
@pytest.mark.parametrize(
    "container,java_version",
    [(OPENJDK_11_CONTAINER, "11")],
    indirect=["container"],
)
```
is that they are not executed when using `-k <marker>` (e.g. `tox -e openjdk -- -n auto -k bci/openjdk_11`) and this is the way we need to execute the validation for single containers. 

So, if I execute `tox -e openjdk -- -n auto -k bci/openjdk_11`, I get failure with `(exited with code 5)`, because no tests have been executed. 
But if I execute `tox -e openjdk -- -n auto` it will execute 1 test `[gw0] [100%] PASSED tests/test_openjdk.py::test_jdk_version[container0-11]` I guess because the marker used here is not `bci/openjdk:11` but `container0-11`, but not sure.

dcermak:
> is that they are not executed when using `-k <marker>` (e.g. `tox -e openjdk -- -n auto -k bci/openjdk_11`) and this is the way we need to execute the validation for single containers.

Ah, I see. The problem here is that this filtering relies on the parameter's marks, which is only taken into account if the container itself is the parameter. What needs to be done would be something like the following (preferably in a loop):
```Python
@pytest.mark.parametrize(
    "container,java_version",
    [pytest.param(OPENJDK_11_CONTAINER, "11", marks=OPENJDK_11_CONTAINER.marks)],
    indirect=["container"],
)
```

tests/test_openjdk_devel.py
```diff
@@ -5,14 +5,28 @@
...
+    [
+        pytest.param(
+            OPENJDK_DEVEL_11_CONTAINER,
+            "11",
+            marks=OPENJDK_DEVEL_11_CONTAINER.marks,
+        ),
+        pytest.param(
+            OPENJDK_DEVEL_17_CONTAINER,
+            "17",
+            marks=OPENJDK_DEVEL_17_CONTAINER.marks,
```
dcermak, line 27: you will need to add an xfail here, see: https://github.com/SUSE/BCI-tests/blob/04b6e72f49b8690b8f54a0a53cd8d7f75b53b4ba/tests/test_openjdk_devel.py#L17

dcermak (approved):

PR #115, opened 2022-05-09
Improve busybox testing

grisu48:
Add two new tests to busybox:

* Check for links in `/bin`
* Test a selection of basic commands for correct behaviour

Related ticket: https://progress.opensuse.org/issues/109812

### DIFF
--- a/tests/test_busybox.py
+++ b/tests/test_busybox.py
@@ -48,10 +48,58 @@ def test_busybox_image_size(
     )
 
 
+def test_busybox_links(auto_container):
+    """Ensure all binaries in :file:`/bin` are links to :file:`/usr/bin/busybox`."""
+    auto_container.connection.run_expect(
+        [0],
+        'for i in /bin/*; do stat -c "%N" "$i" | grep "/usr/bin/busybox"; done',
+    )
+
+
+def test_busybox_binary_works(auto_container):
+    """Ensure the busybox binary works"""
+    busybox = auto_container.connection.run_expect([0], "busybox")
+
+
+def test_true(auto_container):
+    """Test if the busybox `true` and `false` commands are working"""
+    auto_container.connection.run_expect([0], "true")
+    auto_container.connection.run_expect([1], "false")
+
+
+def test_echo_cat_grep_pipes(auto_container):
+    """Test if a string gets passed correctly between `echo`, `cat` and `grep`"""
+    auto_container.connection.run_expect(
+        [0], "echo 'test' | cat | grep 'test'"
+    )
+
+
+def test_ps(auto_container):
+    """Check if the `ps` command yields some output"""
+    assert "root" in auto_container.connection.run_expect([0], "ps").stdout
+
+
+def test_base32_64(auto_container):
+    """Ensure the base32 and base64 commands are returning the correct result for a given "test" string"""
+    assert (
+        "ORSXG5AK"
+        in auto_container.connection.run_expect(
+            [0], "echo test | base32"
+        ).stdout
+    )
+    assert (
+        "dGVzdAo="
+        in auto_container.connection.run_expect(
+            [0], "echo test | base64"
+        ).stdout
+    )
+
+
 @pytest.mark.parametrize(
     "container_per_test", [BUSYBOX_CONTAINER], indirect=True
 )
 def test_busybox_adduser(container_per_test):
+    """Ensure the adduser command works and a new user can be created"""
     container_per_test.connection.run_expect([0], "adduser -D foo")
     getent_passwd = container_per_test.connection.run_expect(
         [0], "getent passwd foo"


tests/test_busybox.py
```diff
@@ -48,6 +48,42 @@ def test_busybox_image_size(
...
+    )
+
+
+@pytest.mark.parametrize(
+    "container_per_test", [BUSYBOX_CONTAINER], indirect=True
+)
+def test_busybox(container_per_test):
+    # Test only a selection of all busybox tools
+    busybox = container_per_test.connection.run_expect([0], "busybox")
+    container_per_test.connection.run_expect([0], "true")
```
jlausuch, line 68: I don't get these checks, but it might be due to my lack of experience with busybox.
How running `true` cmd is special for busybox and not for any random linux environment? `true` will always return 0...
Same for `echo 'test' | cat | grep 'test'` and other checks.. you can run this on any linux env and get the same result, why does it special for busybox?

grisu48: BusyBox is a multi-call binary, an executable program that performs the same job as more than one utility program. That means there is just a single BusyBox binary, but that single binary acts like a large number of utilities. Look at the DESCRIPTION section in https://busybox.net/downloads/BusyBox.html.

All of those program are just symlinks to `/usr/bin/busybox`, see e.g.

    # ls -al /bin
    total 328
    drwxr-xr-x    1 root     root           796 May  8 08:22 .
    drwxr-xr-x    1 root     root           136 May  9 08:18 ..
    lrwxrwxrwx    1 root     root            16 May  8 00:36 arch -> /usr/bin/busybox
    lrwxrwxrwx    1 root     root            16 May  8 00:36 arping -> /usr/bin/busybox
    lrwxrwxrwx    1 root     root            16 May  8 00:36 awk -> /usr/bin/busybox
    ...
    lrwxrwxrwx    1 root     root            16 May  8 00:36 cat -> /usr/bin/busybox
    ...
    lrwxrwxrwx    1 root     root            16 May  8 00:36 true -> /usr/bin/busybox
    ...

In the busybox image, when you run `cat` it does not execute the `/usr/bin/cat` program as standalone, but runs the `busybox` binary and the `cat` applet therein.

The test now runs a bunch of commands to ensure, they are integrated well in the single `busybox` binary by running a subset of those applets and check the expected output, if they are functional and correct.

jlausuch: Yes, that I had some idea, looking at `test_busybox_links` test, however you are using here a small subset, and I was wondering if there is anything special in these commands you are using, and why not using other things like .e.g `ln` or `ls` ?

grisu48: No there is nothing special, those were just low hanging fruit to check if the integration is working properly.

The main point of this PR is to make sure that the system utilities are coming from the `busybox` binary and not from accidentally packaged programs. The rest is just a smoketest to ensure, that `busybox` is working properly.

jlausuch: Ok, as agreed in the chat. In next iteration, we could have the list of expected binaries hardcoded here and make asserts for each one of them to check they exist in the container.

jlausuch (approved):
LGTM

tests/test_busybox.py
```diff
@@ -48,6 +48,42 @@ def test_busybox_image_size(
     )
 
 
+@pytest.mark.parametrize(
+    "container_per_test", [BUSYBOX_CONTAINER], indirect=True
+)
+def test_busybox_links(container_per_test):
+    # Ensure all commands in /bin are links to /usr/bin/busybox
+    container_per_test.connection.run_expect(
```
dcermak, line 56: This is not a mutating test, so no need to use the costly `_per_test` variant.

```suggestion
def test_busybox_links(auto_container):
    """Ensure all binaries in :file:`/bin` are links to :file:`/usr/bin/busybox`."""
    auto_container.connection.run_expect(
```

tests/test_busybox.py
```diff
@@ -48,6 +48,42 @@ def test_busybox_image_size(
...
+        in container_per_test.connection.run_expect(
+            [0], "echo test | base32"
+        ).stdout
+    )
+    assert (
+        "dGVzdAo="
+        in container_per_test.connection.run_expect(
+            [0], "echo test | base64"
+        ).stdout
+    )
```
dcermak, line 84: I would suggest to split this into multiple test functions and give them more descriptive names. Also, none of these tests are performing any mutation, so you can use the `auto_container` fixture here instead.

dcermak (changes requested):
This looks mostly good to me, I have added two suggestions.

tests/test_busybox.py
grisu48: Updated accordingly

tests/test_busybox.py
grisu48: Good point. I moved all subtests to individual tests now. Please mark as resolved if you agree with my changes now.

tests/test_busybox.py
```diff
@@ -48,6 +48,47 @@ def test_busybox_image_size(
...
+
+def test_busybox(auto_container):
+    busybox = auto_container.connection.run_expect([0], "busybox")
+
+
+def test_true(auto_container):
+    auto_container.connection.run_expect([0], "true")
+
+
+def test_echocatgrep(auto_container):
```
dcermak, line 67:
```suggestion
def test_echo_cat_grep_pipes(auto_container):
```

tests/test_busybox.py
```diff
@@ -48,6 +48,47 @@ def test_busybox_image_size(
...
 
+def test_busybox_links(auto_container):
+    """Ensure all binaries in :file:`/bin` are links to :file:`/usr/bin/busybox`."""
+    auto_container.connection.run_expect(
+        [0],
+        'for i in /bin/*; do stat -c "%N" "$i" | grep "/usr/bin/busybox"; done',
+    )
+
+
+def test_busybox(auto_container):
```
dcermak, line 59:
```suggestion
def test_busybox_binary_works(auto_container):
```

tests/test_busybox.py
```diff
@@ -48,6 +48,47 @@ def test_busybox_image_size(
...
+    busybox = auto_container.connection.run_expect([0], "busybox")
+
+
+def test_true(auto_container):
+    auto_container.connection.run_expect([0], "true")
```
dcermak, line 64: How about testing false as well?
```suggestion
def test_true_false_binaries(auto_container):
    auto_container.connection.run_expect([0], "true")
    auto_container.connection.run_expect([1], "false")
```

dcermak (changes requested):
Would you please add docstrings to every test function? I know it's tedious, but this will make sure that the test show up in the written documentation as well.

tests/test_busybox.py
grisu48: done

tests/test_busybox.py
grisu48: done

tests/test_busybox.py
grisu48: done

jlausuch (approved):
Looks much better now :)

dcermak (approved):

PR #117, opened 2022-05-11
Use per test fixtures with 389 ds

dcermak:
This is required as https://github.com/dcermak/pytest_container/pull/81 has now been merged

### DIFF
--- a/tests/test_all.py
+++ b/tests/test_all.py
@@ -42,77 +42,87 @@ CMD ["/fetcher/main"]
 """
 
 
-def test_os_release(auto_container):
+def test_os_release(auto_container_per_test):
     """
     :file:`/etc/os-release` is present and the values of ``OS_VERSION`` and
     ``OS_PRETTY_NAME`` equal :py:const:`bci_tester.data.OS_VERSION` and
     :py:const:`bci_tester.data.OS_PRETTY_NAME` respectively
     """
-    assert auto_container.connection.file("/etc/os-release").exists
+    assert auto_container_per_test.connection.file("/etc/os-release").exists
 
     for (var_name, value) in (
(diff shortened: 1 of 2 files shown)


tests/test_all.py
```diff
@@ -42,77 +42,87 @@
 """
 
 
-def test_os_release(auto_container):
+def test_os_release(auto_container_per_test):
```
grisu48, line 45: Is a conter per test for checking `/etc/os-release` really necessary? After a quick look at https://github.com/dcermak/pytest_container/pull/81 I don't see why this is needed now but I might be missing something. `auto_container` should here also work, no?

dcermak:
Felix Niederwanger ***@***.***> writes:

> @grisu48 commented on this pull request.
>
>
>
>> @@ -42,77 +42,87 @@
>  """
>  
>  
> -def test_os_release(auto_container):
> +def test_os_release(auto_container_per_test):
>
> Is a conter per test for checking `/etc/os-release` really necessary? After a quick look at https://github.com/dcermak/pytest_container/pull/81 I don't see why this is needed now but I might be missing something. `auto_container` should here also work, no?

The problem is the 389-ds container: it is set as `singleton=True` which
means that it can only be used in a `_per_test`
fixture. pytest_container will now throw an error otherwise as you can
run into deadlocks if you use it with a auto_container or container
fixture (that is what we've been debugging with Jose for the past week…).

tests/test_all.py
grisu48: https://github.com/SUSE/BCI-tests/pull/117#issuecomment-1123548255

grisu48: I see, thanks for the clarification. That was not immediately visible from the two PRs to me.

PR #119, opened 2022-05-12
Split minimal and micro tests

grisu48:
Split the tests of BCI micro and BCI minimal into two different files
for better overview and maintainability.

Related ticket: https://progress.opensuse.org/issues/109431

### DIFF
--- a/tests/test_minimal.py
+++ b/tests/test_minimal.py
@@ -3,9 +3,11 @@ from typing import Dict
 import pytest
 from pytest_container.runtime import LOCALHOST
 
-from bci_tester.data import MICRO_CONTAINER
 from bci_tester.data import MINIMAL_CONTAINER
 
+CONTAINER_IMAGES = [
+    MINIMAL_CONTAINER,
+]
 
 #: size limits of the minimal image per architecture in MiB
 MINIMAL_IMAGE_MAX_SIZE: Dict[str, int] = {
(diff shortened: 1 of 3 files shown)


tests/test_minimal.py
```diff
@@ -45,9 +33,7 @@ def test_minimal_image_size(
     )
 
 
-@pytest.mark.parametrize(
-    "container", [MICRO_CONTAINER, MINIMAL_CONTAINER], indirect=True
-)
+@pytest.mark.parametrize("container", [MINIMAL_CONTAINER], indirect=True)
```
jlausuch, line 36: do we need this ? can't we just use `auto_container`?

grisu48: If I use `auto_container` I get the following error message:

    E   ValueError: The test function test_fat_packages_absent is using the auto_container fixture but the parent module is not setting the 'CONTAINER_IMAGES' variable

Can somebody point me in the direction on how to solve such issues?

jlausuch: Yes, you would need to define the constant:
```
CONTAINER_IMAGES = [
    MINIMAL_CONTAINER,
]
```
e.g. https://github.com/SUSE/BCI-tests/blob/main/tests/test_python.py#L7

grisu48: thnx and done

dcermak:
How does this improve maintainability? We now have just more duplication…

jlausuch:
> How does this improve maintainability? We now have just more duplication…

The way this suite is used in production is by specifying the environment to the tox command, so executing one of the modules under /tests/. 
Micro and Minimal are tested and released separately, so having different test suites for each of them makes things more logical.

For code duplication, like the fat packages check, that can be solved by moving that method to `test_all.py` and make sure it runs only for the desired containers. Something similar to https://github.com/SUSE/BCI-tests/blob/main/tests/test_all.py#L107

dcermak:
> The way this suite is used in production is by specifying the environment to the tox command, so executing one of the modules under /tests/.
> Micro and Minimal are tested and released separately, so having different test suites for each of them makes things more logical.

What would block you from using the same filtering logic that you are using for all?

pdostal (approved):

grisu48:
Closing as this is not needed anymore.

PR #122, opened 2022-05-13
Add pcp tests

djoreilly:
This follows the init tests, as the pcp container is based on the
init container. The docker test is removed as only podman is supported.

WIP
export TARGET=x
export BASEURL=registry.suse.de/home/doreilly/branches/suse/sle-15-sp4/update/bci
tox -e pcp

### DIFF
--- a/bci_tester/data.py
+++ b/bci_tester/data.py
@@ -311,7 +311,8 @@ INIT_CONTAINER = create_BCI(
 PCP_CONTAINER = create_BCI(
     build_tag=f"suse/pcp:5.2.2",
     image_type="dockerfile",
-    available_versions=["15.3"],
+    singleton=True,
+    extra_marks=[pytest.mark.skipif(DOCKER_SELECTED, reason="only podman is supported")],
     extra_launch_args=[]
     if DOCKER_SELECTED
     else [
--- a/bci_tester/data.py
+++ b/bci_tester/data.py
@@ -308,6 +308,21 @@ INIT_CONTAINER = create_BCI(
     default_entry_point=True,
 )
 
+PCP_CONTAINER = create_BCI(
+    build_tag=f"suse/pcp:5.2.2",
+    image_type="dockerfile",
+    available_versions=["15.3"],
+    extra_launch_args=[]
+    if DOCKER_SELECTED
+    else [
+        "--systemd",
+        "always",
+        "-p",
+        "44322:44322",
+    ],
+    default_entry_point=True,
+)
+
 CONTAINER_389DS = create_BCI(
     build_tag="suse/389-ds:2.0",
     image_type="dockerfile",
@@ -358,6 +373,7 @@ CONTAINERS_WITH_ZYPPER = [
     NODEJS_12_CONTAINER,
     NODEJS_14_CONTAINER,
     NODEJS_16_CONTAINER,
+    PCP_CONTAINER,
     PYTHON36_CONTAINER,
     PYTHON39_CONTAINER,
     PYTHON310_CONTAINER,
--- a/bci_tester/data.py
+++ b/bci_tester/data.py
@@ -312,7 +312,9 @@ PCP_CONTAINER = create_BCI(
     build_tag=f"suse/pcp:5.2.2",
     image_type="dockerfile",
     singleton=True,
-    extra_marks=[pytest.mark.skipif(DOCKER_SELECTED, reason="only podman is supported")],
+    extra_marks=[
+        pytest.mark.skipif(DOCKER_SELECTED, reason="only podman is supported")
+    ],
     extra_launch_args=[]
     if DOCKER_SELECTED
     else [
--- a/tests/test_pcp.py
+++ b/tests/test_pcp.py
@@ -16,27 +16,28 @@ def test_systemd_status(auto_container_per_test):
 def test_pcp_services_status(auto_container_per_test):
     """Check that the pcp services are healthy."""
 
-    wait_for_pmcd(auto_container_per_test)
-
-    auto_container_per_test.connection.run_expect([0], "systemctl status pmcd")
-    auto_container_per_test.connection.run_expect(
-        [0], "systemctl status pmlogger"
-    )
-    auto_container_per_test.connection.run_expect(
-        [0], "systemctl status pmproxy"
-    )
-    auto_container_per_test.connection.run_expect([0], "systemctl status pmie")
+    for service in ("pmcd", "pmlogger", "pmproxy", "pmie"):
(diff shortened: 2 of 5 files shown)


jlausuch:
This is great start!
I would actually split the function you created into different tests. After our chat, I put together a test plan, at least to start with some smoke test: https://confluence.suse.com/display/qasle/PCP+Container+testing

Basically, what I suggest is having separated functions (sub-tests) for each case:
```
1) Check systemd present (similar to https://github.com/SUSE/BCI-tests/blob/main/tests/test_init.py#L18)

2) Check that the 4 daemons are running

    pmcd
    pmlogger
    pmproxy
    pmie

3) Run some basic pminfo commands, for example:

    podman exec pcp pminfo --fetch disk.dev.write
   podman exec pcp pmprobe -v mem.physmem


4) Check service from the host
      curl -s http://localhost:44322/metrics?names=mem.physmem
should return the same as
   podman exec pcp pminfo --fetch mem.physmem
```

tests/test_pcp.py
```diff
@@ -0,0 +1,37 @@
...
+
+    # test call to pmcd
+    assert container.connection.run_expect([0], "pmprobe -v mem.physmem")
+
+    # test call to pmproxy
+    if LOCALHOST.exists("curl"):
+        assert (
+            LOCALHOST.run_expect(
+                [0],
+                "curl -s http://localhost:44322/metrics?names=mem.physmem",
```
dcermak, line 35: Beware that this requires curl to be installed on the host system. If you do nothing with the data, then you could also just check that the container has this port open.

tests/test_pcp.py
```diff
@@ -0,0 +1,37 @@
...
+
+@pytest.mark.parametrize("container", [PCP_CONTAINER], indirect=True)
+@pytest.mark.skipif(DOCKER_SELECTED, reason="only podman is supported")
+def test_systemd_present(container):
+    """Check that the pcp daemons are running.
+
+    """
+
+    # pcp needs a little time to initialize
+    time.sleep(5)
```
dcermak, line 19: Beware, this will break on slow hosts. Have you considered using a HealthCheck instead? pytest_container supports waiting until the container is healthy.

bci_tester/data.py
```diff
@@ -304,6 +304,21 @@ def create_BCI(
...
+    extra_launch_args=[]
+    if DOCKER_SELECTED
+    else [
+        "--systemd",
+        "always",
+        "-p",
+        "44322:44322",
+    ],
+    default_entry_point=True,
+)
```
dcermak, line 320: I would suggest that you make this container a singleton as you must not launch two instances of it.
Also, I would suggest that you move the skipIf from test_pcp into this call via the `extra_marks` parameter.

```suggestion
PCP_CONTAINER = create_BCI(
    build_tag=f"suse/pcp:5.2.2",
    image_type="dockerfile",
    singleton=True,
    extra_marks=[pytest.mark.skipif(DOCKER_SELECTED, reason="only podman is supported")],
    extra_launch_args=[
        "--systemd",
        "always",
        "-p",
        "44322:44322",
    ],
    default_entry_point=True,
)
```

dcermak (commented):
Very nice, thank you!

tests/test_pcp.py
jlausuch: `curl` is  pre-installed in the container hosts of our tests, so no issue here.

tests/test_pcp.py
djoreilly: nope, this was just a quick hack to make it pass. I'll look into the HealthCheck - thanks

tests/test_pcp.py
djoreilly: I just followed the example from test_389ds.py. I'm thinking of writing a helper function to get MemTotal from `/proc/meminfo` to assert against

tests/test_pcp.py
djoreilly: Do you mean `healthcheck_timeout_ms`? That waits until the container is started?  But here we need to wait until pcp services are running (not initializing). Also with docker it seems you can give your container image a healthcheck command, but with podman you can only to provide it on run.

tests/test_pcp.py
```diff
@@ -0,0 +1,56 @@
...
+
+
+def test_systemd_status(auto_container_per_test):
+    auto_container_per_test.connection.run_expect([0], "systemctl status")
+
+
+def test_pcp_services_status(auto_container_per_test):
+    """Check that the pcp services are healthy."""
+
+    wait_for_pmcd(auto_container_per_test)
```
jlausuch, line 19: where is this function defined?

tests/test_pcp.py
```diff
@@ -0,0 +1,56 @@
...
+        assert LOCALHOST.run_expect(
+            [0],
+            "curl -s http://localhost:44322/metrics?names=mem.physmem",
+        )
+
+
+def wait_for_pmcd(con):
+    """pmcd takes a little time to initialize things before it is ready"""
+
+    for _ in range(30):
```
jlausuch, line 50: I would actually increase this timeout. We are running the tests on multiple architectures and in multiple host versions. This is an example of a timeout I had to increase: https://github.com/SUSE/BCI-tests/pull/121/files

tests/test_pcp.py
```diff
@@ -0,0 +1,56 @@
...
+        rc = con.connection.run("systemctl status pmcd").rc
```
jlausuch, line 51: I would actually use something that returns true if it's really running, like `systemctl is-active pmcd`

jlausuch:
I am missing the test that checks the 4 daemons: `pmcd` `pmlogger`  `pmproxy` `pmie`.
There is that wait function checking for `pmcd`, but is it enough? Should we have a specific test for the other daemons?

tests/test_pcp.py
djoreilly: will do

jlausuch:
@djoreilly is this ready o are you planning to amend it? Maybe we could merge and have the container enabled already in the test pipeline and extend the coverage later on?

djoreilly:
> @djoreilly is this ready o are you planning to amend it? Maybe we could merge and have the container enabled already in the test pipeline and extend the coverage later on?

@jlausuch okay I don't have anything else.

jlausuch (approved):

jlausuch:
I'm merging and enabling the pcp container in the pipleine. Let's handle next issues in future PRs. Thanks a lot!

jlausuch:
I created some small fixes: https://github.com/SUSE/BCI-tests/pull/129

PR #123, opened 2022-05-13
Architecture not x86 skip tensorflow test

m-dati:
The test https://openqa.suse.de/tests/8754641 for aarch64 architecture passed all python tests, but not that on Tensorflow, because a ready to work library is actually not available for architectures other than x86-64, so in such cases this test will be skipped.

### DIFF
--- a/tests/test_python.py
+++ b/tests/test_python.py
@@ -2,6 +2,7 @@
 import pytest
 from pytest_container import DerivedContainer
 from pytest_container.container import container_from_pytest_param
+from pytest_container.runtime import LOCALHOST
 
 from bci_tester.data import PYTHON310_CONTAINER
 from bci_tester.data import PYTHON36_CONTAINER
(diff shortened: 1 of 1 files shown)


m-dati:
@dcermak , @jlausuch, Please a kind check and approval for this merge request.

jlausuch (approved):
Yes, after some tests in openQA, we can only run this test on x86_64. I have also adjusted the openQA tests to use SSE4 cpu flag

jlausuch:
Please run `tox -e format`

jlausuch:
Looks like it fails now with `ValueError: invalid literal for int() with base 10: 'T'`
https://openqa.suse.de/tests/8768094#step/python/6

dcermak:
> Looks like it fails now with `ValueError: invalid literal for int() with base 10: 'T'` https://openqa.suse.de/tests/8768094#step/python/6

Is that somehow reproducible? Because I cannot reproduce this…

jlausuch:
> > Looks like it fails now with `ValueError: invalid literal for int() with base 10: 'T'` https://openqa.suse.de/tests/8768094#step/python/6
> 
> Is that somehow reproducible? Because I cannot reproduce this…

Yes, only with python 3.6 and 3.9 images... https://openqa.suse.de/tests/overview?distri=sle&version=15-SP3&build=_15-SP3_17.15_python-3.9-image&groupid=409

PR #124, opened 2022-05-17
WIP: Fix race condition in python_webserver_1

grisu48:
Fix a race condition in the python_webserver_1 test case, where the test
fails because the http webserver is not yet ready when being checked to
be present.

Related ticket: https://progress.opensuse.org/issues/111164 
Verification run: [failing without fix](https://duck-norris.qam.suse.de/tests/8895#) | [passing with fix](https://duck-norris.qam.suse.de/tests/8905)

### DIFF
--- a/tests/test_python.py
+++ b/tests/test_python.py
@@ -1,5 +1,6 @@
 """Basic tests for the Python base container images."""
 import pytest
+import time
 from pytest_container import DerivedContainer
 from pytest_container.container import container_from_pytest_param
 from pytest_container.runtime import LOCALHOST
@@ -96,17 +97,23 @@ def test_python_webserver_1(container_per_test):
     ).is_listening
 
     # start of the python http server
-    bci_pyt_serv = container_per_test.connection.run_expect(
-        [0], f"timeout 240s python3 -m http.server {port} &"
-    ).stdout
+    container_per_test.connection.run_expect(
+        [0], f"python3 -m http.server {port} &"
+    )
+
+    # wait until the python http.server process is running in the container
+    timeout = 30
+    for i in range(timeout + 1):
+        assert i < timeout, "webserver startup timeout"
+        if container_per_test.connection.socket(
+            "tcp://0.0.0.0:" + port
+        ).is_listening:
+            break
+        time.sleep(1)
 
     # checks that the python http.server process is running in the container:
     proc = container_per_test.connection.process.filter(comm="python3")
-
-    # check that the filtered list is not empty
     assert len(proc) > 0, "The python3 http.server process must be running"
-
-    x = None
 
     for p in proc:
         x = p.args
@@ -115,11 +122,6 @@ def test_python_webserver_1(container_per_test):
 
     # checks expected parameter of the running python process
     assert "http.server" in x, "http.server not running."
-
-    # checks that the expected port is listening in the container
-    assert container_per_test.connection.socket(
-        "tcp://0.0.0.0:" + port
-    ).is_listening, "Error on the expected port"
 
 
 @pytest.mark.parametrize(


tests/test_python.py
```diff
@@ -96,9 +97,12 @@ def test_python_webserver_1(container_per_test):
...
 
     # start of the python http server
-    bci_pyt_serv = container_per_test.connection.run_expect(
+    container_per_test.connection.run_expect(
         [0], f"timeout 240s python3 -m http.server {port} &"
-    ).stdout
+    )
+
+    # give the http.server some time to start
+    time.sleep(5)
```
jlausuch, line 105: what about a timeout of e.g. 10s or more with a loop doing `nc -vz {port}`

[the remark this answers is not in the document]
pdostal: Please avoid sleep.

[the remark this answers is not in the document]
grisu48: Please provide a suitable alternative suggestion on how we can avoid this sleep here.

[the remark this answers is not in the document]
jlausuch: Maybe not the best way to implement it, but the idea is something like
```
rc = 1
for _ in range(60):
    rc = con.connection.run("nc -z localhost {port}").rc:
    if rc == 0:
        break
    time.sleep(1)
    
if rc != 0 -> raise error    
```

tests/test_python.py
```diff
@@ -96,9 +97,16 @@ def test_python_webserver_1(container_per_test):
...
+    container_per_test.connection.run_expect(
         [0], f"timeout 240s python3 -m http.server {port} &"
-    ).stdout
+    )
+
+    # wait for the webserver to start
+    timeout = 60
+    for i in range(timeout + 1):
+        assert i < timeout, "webserver startup timeout"
+        if container_per_test.connection.run("nc -z localhost {port}").rc:
```
jlausuch, line 108: not sure if we need to use `f""` here, as this might consider `{port}` as a string and not a variable. Have you tried the code?

jlausuch: also confusing, you are using `import time` but you don't use it. The way it's now, you need a `sleep(1)` somewhere, oder?

grisu48: Working on it right now.

tests/test_python.py
```diff
@@ -96,9 +97,17 @@ def test_python_webserver_1(container_per_test):
...
-    ).stdout
+    )
+
+    # wait for the webserver to start
+    timeout = 60
+    for i in range(timeout + 1):
+        assert i < timeout, "webserver startup timeout"
+        if container_per_test.connection.run(f"nc -z localhost {port}").rc:
+            break
+        time.sleep(1)
```
jlausuch, line 110: and `i++`?

tests/test_python.py
```diff
@@ -1,5 +1,6 @@
 """Basic tests for the Python base container images."""
 import pytest
+import time
```
jlausuch, line 3: do you need this?

grisu48: `time.sleep` in line 111

tests/test_python.py
grisu48: `for i in range(timeout + 1):` in line 106

tests/test_python.py
```diff
@@ -96,18 +97,24 @@ def test_python_webserver_1(container_per_test):
     ).is_listening
 
     # start of the python http server
-    bci_pyt_serv = container_per_test.connection.run_expect(
-        [0], f"timeout 240s python3 -m http.server {port} &"
-    ).stdout
+    container_per_test.connection.run_expect(
+        [0], f"python3 -m http.server {port} &"
```
dcermak, line 101: Another option would be to modify the entrypoint of `CONTAINER_IMAGES_T` to `["/usr/bin/python3",  "-m", "http.server"]` and add a `HEALTHCHECK` that will check whether the server is running. Then `pytest_container` will do the waiting for you.

grisu48:
I guess this issue has been resolved in the meantime and I'm gonna close this PR.

PR #132, opened 2022-06-06
Bci python webserver_1 fix for podman old version

m-dati:
an Issue was found in the test_python - test_webserver_1 that was related to the hdd_1 qcow2 having an old podman version 1.x.x.
Here the VR:
http://marvin5.arch.suse.cz/tests/229  # sles es 8.5, regression podman.v.4x: `PASS`
http://marvin5.arch.suse.cz/tests/228  # sles es 7.9, fix podman.v.1x: `PASS`!

### DIFF
--- a/tests/test_python.py
+++ b/tests/test_python.py
@@ -1,7 +1,8 @@
 """Basic tests for the Python base container images."""
+import pytest
+import re
 import time
 
-import pytest
 from bci_tester.data import PYTHON310_CONTAINER
 from bci_tester.data import PYTHON36_CONTAINER
 from bci_tester.data import PYTHON39_CONTAINER
@@ -16,6 +17,8 @@ outdir = "output/"
 appl1 = "tensorflow_examples.py"
 port1 = 8123
 t0 = time.time()
+
+#container_version = lambda container_runtime: re.sub("[^0-9.:_-]", "", LOCALHOST.run(container_runtime.runner_binary + " --version")
 
 # copy tensorflow module trainer from the local application directory to the container
 DOCKERF_PY_T1 = f"""
@@ -46,7 +49,7 @@ CONTAINER_IMAGES_T1 = [
         DerivedContainer(
             base=container_from_pytest_param(CONTAINER_T),
             containerfile=DOCKERF_PY_T1,
-            extra_launch_args=["-d", "-p", f"{port1}:{port1}"],
+            extra_launch_args=["-p", f"{port1}:{port1}"],
         ),
         marks=CONTAINER_T.marks,
         id=CONTAINER_T.id,
@@ -67,6 +70,8 @@ CONTAINER_IMAGES_T2 = [
     for CONTAINER_T in CONTAINER_IMAGES
 ]
 
+# get container version and clean the result from not (numeric or separators) chars.
+container_version = lambda container_runtime: re.sub("[^0-9.,:_-]", "", LOCALHOST.run(container_runtime.runner_binary + " --version").stdout)
 
 def test_python_version(auto_container):
     """Test that the python version equals the value from the environment variable
@@ -108,21 +113,28 @@ def test_tox(auto_container):
     "container_per_test", CONTAINER_IMAGES_T1, indirect=["container_per_test"]
 )
 @pytest.mark.parametrize("hmodule, port, retry", [("http.server", port1, 10)])
-def test_python_webserver_1(container_per_test, hmodule, port, retry):
+def test_python_webserver_1(
+    container_per_test, container_runtime, hmodule, port, retry
+):
     """Test that the python webserver is able to open a given port"""
 
     portstatus = False
-
+    podman_version_x = None
     t = 0
+
+    if container_runtime.runner_binary == "podman":
+        podman_version = container_version(container_runtime).split(".")
+        assert podman_version[0].isnumeric()
+
+        # container version: old versions have issues with background processes
+        if int(podman_version[0]) < 2:
+            pytest.xfail("server port checks not compatible with old podman versions 1")
+
+    command = f"timeout --preserve-status 120 python3 -m {hmodule} {port} &"
 
     # pkg neeed to run socket/port check
     if not container_per_test.connection.package("iproute2").is_installed:
         container_per_test.connection.run_expect([0], "zypper -n in iproute2")
-
-    #p2=LOCALHOST.socket(f"tcp://127.0.0.1:{port}").is_listening
-    # p3=LOCALHOST.run_expect([0], f"curl http://127.0.0.1:{port}")
-    # p3=LOCALHOST.run(f"timeout --preserve-status 30 wget http://127.0.0.1:{port}/{appl1}")
-    #print("ante start", p2) #, p3.stdout)
 
     # checks that the expected port is Not listening yet
     assert not container_per_test.connection.socket(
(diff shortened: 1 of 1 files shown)


tests/test_python.py
```diff
@@ -106,13 +107,29 @@ def test_tox(auto_container):
     "container_per_test", CONTAINER_IMAGES_T1, indirect=["container_per_test"]
 )
 @pytest.mark.parametrize("hmodule, port, retry", [("http.server", port1, 10)])
```
dcermak, line 109: Why is this parametrized?

tests/test_python.py
```diff
@@ -106,13 +107,29 @@ def test_tox(auto_container):
...
-def test_python_webserver_1(container_per_test, hmodule, port, retry):
+def test_python_webserver_1(
+    container_per_test, container_runtime, hmodule, port, retry
+):
     """Test that the python webserver is able to open a given port"""
 
+    check = ""
```
dcermak, line 115: Please use a more descriptive variable name.

tests/test_python.py
```diff
@@ -106,13 +107,29 @@ def test_tox(auto_container):
...
     t = 0
 
+    if container_runtime.runner_binary == "podman":
+        vv = LOCALHOST.package(container_runtime.runner_binary).version
+        vn = vv.split(".")
+        assert vn[0].isnumeric()
+
+        # container version: old versions needs a containerized check of background processes
+        if int(vn[0]) < 4:
+            check = f"sleep 1; ss --numeric --listening --tcp | grep {port};"
```
dcermak, line 127: Please elaborate in more detail why you are doing the following with podman < 4.0. Also, if you need to check for open sockets, please use testinfra for that: https://testinfra.readthedocs.io/en/latest/modules.html?highlight=port#testinfra.modules.socket.Socket
Adding these shell scripts everywhere will turn into a maintenance nightmare in the long run.

tests/test_python.py
```diff
@@ -125,25 +142,32 @@ def test_python_webserver_1(container_per_test, hmodule, port, retry):
...
-    for t in range(retry):
-        time.sleep(1)
-        portstatus = container_per_test.connection.socket(
-            f"tcp://0.0.0.0:{port}"
-        ).is_listening
-        if portstatus:
-            break
+    if len(check) > 0:
+        portstatus = str(port) in server.stdout
+        # retry = 0 # skip the loop
```
dcermak, line 152:
```suggestion
```

tests/test_python.py
```diff
@@ -44,7 +45,7 @@
         DerivedContainer(
             base=container_from_pytest_param(CONTAINER_T),
             containerfile=DOCKERF_PY_T1,
-            extra_launch_args=["-p", f"{port1}:{port1}"],
+            extra_launch_args=["-d", "-p", f"{port1}:{port1}"],
```
dcermak, line 48: Why is this necessary? `pytest_container` will launch your container in the background for you, you shouldn't have to add a deamonize flag to it.

tests/test_python.py
```diff
@@ -125,25 +142,32 @@ def test_python_webserver_1(container_per_test, hmodule, port, retry):
...
+
+    x = container_per_test.connection.run("ps ax | grep http|grep -v grep")
 
     t2 = time.time() - t0
 
     # check inspection success or timeout
     assert (
         portstatus
-    ), f"Timeout expired:Before start {t1}s After {t} checks {t2}s. Expected port not listening"
+    ), f"Timeout expired:Before start {t1}-{t1b}s After {t} checks {t2}s. Expected port not listening"
```
dcermak, line 170: What do you mean by "before N seconds After M checks X seconds."? I don't understand the sentence.

dcermak (changes requested):
I must confess that I am a little afraid into what this module is turning…

Also, the CI fails on github:
```ShellSession
>           assert vn[0].isnumeric()
E           AssertionError: assert False
E            +  where False = <built-in method isnumeric of str object at 0x7f2133598870>()
E            +    where <built-in method isnumeric of str object at 0x7f2133598870> = '100:3'.isnumeric
```

tests/test_python.py
```diff
@@ -106,13 +107,29 @@ def test_tox(auto_container):
...
+        vn = vv.split(".")
+        assert vn[0].isnumeric()
+
+        # container version: old versions needs a containerized check of background processes
+        if int(vn[0]) < 4:
+            check = f"sleep 1; ss --numeric --listening --tcp | grep {port};"
+
+    command = f"(timeout --preserve-status 30 nohup python3 -m {hmodule} {port} &);{check}"
+
+    print("cv version:", vv, command)
```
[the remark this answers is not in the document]
m-dati: About this test, it passed in a previous merge, but in the last weeks runs it showed some unexpected differences of behavior for Podman version 1.x, so I added the related package check.
 
When it comes to the "isnumeric" assertion, it is needed to intercept the eventual other version formats different from the expected `x.y.z `

And in fact the CI error here is due to the **testinfra** choice to use [dpkg-query](https://github.com/SUSE/BCI-tests/runs/6759533826?check_suite_focus=true#step:6:131) for this run, that returns `stdout=b'install ok installed 100:3.4.2-4`' , where "**100:**" is not part of podman [version](https://github.com/SUSE/BCI-tests/runs/6759533826?check_suite_focus=true#logs)
So eventually in podman_version a more string analysis is needed.

m-dati:
The issue is that the python3 process started 

> > > What about having `skip` if podman version < 2.x.y?
> > > This issue only happens on a host with podman `1.6.4`
> > 
> > 
> > That sounds like a much more reasonable suggestion, unless we really **must** run this test with an ancient podman version (is that version even _supported_ anywhere??)
> 
> The idea was originally from Maurizio, and I think I would vote for it rather than this PR. 1.6.4 is default in Expanded Support 7.9 (RHEL).

To skip the test for a given version, the version itself shall be better defined, as I described [above](https://github.com/SUSE/BCI-tests/pull/132#discussion_r891196785)

I will try to improve that version check and apply the skip for v<2.x.y

tests/test_python.py
m-dati: To improve the parameter visibility and manage possible future change for test reuse or debug.

tests/test_python.py
m-dati: sentence changed in the new commit

tests/test_python.py
m-dati: removed

tests/test_python.py
```diff
@@ -106,13 +107,29 @@ def test_tox(auto_container):
...
-def test_python_webserver_1(container_per_test, hmodule, port, retry):
+def test_python_webserver_1(
+    container_per_test, container_runtime, hmodule, port, retry
+):
     """Test that the python webserver is able to open a given port"""
 
+    check = ""
     portstatus = False
-
+    vv = None
```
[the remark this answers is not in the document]
m-dati: changed in new commit

tests/test_python.py
```diff
@@ -125,25 +142,32 @@ def test_python_webserver_1(container_per_test, hmodule, port, retry):
...
+        for t in range(retry):
+            time.sleep(1)
+            portstatus = container_per_test.connection.socket(
+                f"tcp://0.0.0.0:{port}"
+            ).is_listening
+            print(t, time.time() - t0)
+            if portstatus:
+                break
+
+    x = container_per_test.connection.run("ps ax | grep http|grep -v grep")
```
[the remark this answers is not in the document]
m-dati: this command was for previous debugging, removed.

tests/test_python.py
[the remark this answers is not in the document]
m-dati: As discussed, I provided a version check and test skip in case of 1.x

tests/test_python.py
m-dati: After many check and attempts for debugging a failing test, I noted that after stared the python http server, in case of podman v.1.x any next testinfra command remains **waiting** for that process to exit after timeout expired, before to go on, even being in detach mode and background. Therefore those testinfra port/process checks will have that server no more running. This case does not happens with the newer versions of podman.

tests/test_python.py
```diff
@@ -106,13 +115,27 @@ def test_tox(auto_container):
...
 
+    if container_runtime.runner_binary == "podman":
+        podman_version = container_version(container_runtime).split(".")
+        assert podman_version[0].isnumeric()
+
+        # container version: old versions have issues with background processes
+        if int(podman_version[0]) < 2:
+            pytest.xfail(
+                "server port checks not compatible with old podman versions 1"
+            )
```
dcermak, line 135: Please skip the whole test, there's really no need to run it if we know that it is broken

tests/test_python.py
```diff
@@ -65,6 +67,13 @@
     for CONTAINER_T in CONTAINER_IMAGES
 ]
 
+# get container version and clean the result from not (numeric or separators) chars.
+container_version = lambda container_runtime: re.sub(
+    "[^0-9.,:_-]",
+    "",
+    LOCALHOST.run(container_runtime.runner_binary + " --version").stdout,
+)
```
dcermak, line 75:
```suggestion
def get_container_runtime_version(container_runtime: OciRuntimeBase) -> str:
    """get the container runtime version and clean the result from non-numeric or non-separators characters."""
    return re.sub(
        "[^0-9.,:_-]",
        "",
        LOCALHOST.run(container_runtime.runner_binary + " --version").stdout,
    )
```

tests/test_python.py
```diff
@@ -19,6 +20,7 @@
 # copy tensorflow module trainer from the local application directory to the container
 DOCKERF_PY_T1 = f"""
 WORKDIR {bcdir}
+COPY {orig + appdir}/{appl1}  {bcdir}
```
dcermak, line 23: Why is this change necessary?

m-dati:
BCI python webserver_1 for podman version 1, in the following test: 
http://marvin5.arch.suse.cz/tests/277
The full run pass now,but skipping that specific test.

In this other test, the webserver_1 pass instead, being podman v 4.x:
http://marvin5.arch.suse.cz/tests/278

dcermak (commented):
Looks pretty good, I have added a small suggestion and have one question left.

tests/test_python.py
m-dati: Used in a previous debugging, now no more needed,it will be removed

tests/test_python.py
m-dati: applied

m-dati:
Probably, in case of a wider need, in some future review, the container version function could be moved to pytest_container.runtime, importing the value as a global parameter.

dcermak:
@m-dati please rebase your branch

dcermak:
@m-dati reformat the source code please, the `format.sh` script was broken

dcermak (approved):
LGTM, thanks!

m-dati:
Many thanks to you too

dcermak:
> @dcermak Would it be possible to have some sort of new variable `PODMAN_VERSION` or `ENGINE_VERSION` in pytest to avoid `get_container_runtime_version(get_selected_runtime()).split(".")[0]` ?

WDYT: https://github.com/dcermak/pytest_container/pull/91 ?

PR #139, opened 2022-06-10
Improve bci-init tests

grisu48:
Improve the systemd tests for the bci-init container by adding multiple
new tests for boottime, journald, timedatectl, loginctl, hostnamectl.

Related ticket: https://progress.opensuse.org/issues/111054

### DIFF
--- a/tests/test_init.py
+++ b/tests/test_init.py
@@ -2,27 +2,185 @@
 systemd pre-installed.
 
 """
+import datetime
+from typing import Dict
+
 import pytest
 from bci_tester.data import INIT_CONTAINER
 from bci_tester.runtime_choice import DOCKER_SELECTED
 from pytest_container.runtime import LOCALHOST
 
+CONTAINER_IMAGES = [INIT_CONTAINER]
 
-@pytest.mark.parametrize("container", [INIT_CONTAINER], indirect=True)
+
 @pytest.mark.skipif(
     DOCKER_SELECTED
     and (int(LOCALHOST.package("systemd").version.split(".")[0]) >= 248),
     reason="Running systemd in docker is broken as of systemd 248, see https://github.com/moby/moby/issues/42275",
 )
-def test_systemd_present(container):
-    """Check that :command:`systemctl` is in ``$PATH``, that :command:`systemctl
-    status` works and that :file:`/etc/machine-id` exists.
-
-    This test is currently broken due to `moby/moby#42275
-    <https://github.com/moby/moby/issues/42275>`_ with :command:`docker` and
-    :command:`systemd` >= 248.
-
+class TestSystemd:
     """
-    assert container.connection.exists("systemctl")
-    assert container.connection.file("/etc/machine-id").exists
-    assert container.connection.run_expect([0], "systemctl status")
+    systemd test module for the bci-init container.
+    """
+
+    def test_systemd_present(self, auto_container):
+        """Check that :command:`systemctl` is in ``$PATH``, that :command:`systemctl
+        status` works and that :file:`/etc/machine-id` exists.
+        """
+        assert auto_container.connection.exists("systemctl")
+        assert auto_container.connection.file("/etc/machine-id").exists
+        assert auto_container.connection.run_expect([0], "systemctl status")
+
+    def test_systemd_boottime(self, auto_container):
+        """Ensure the container startup time is below 5 seconds"""
+
+        STARTUP_LIMIT = 5  # Container startup time limit in seconds
+        TARGET_LIMIT = (
+            5  # Limit for the systemd target to be reached in seconds
+        )
+
+        def extract_time(stdout, prefix):
+            """internal helper function to extract the time from systemd-analyze time
+            extracts a timestamp with a given prefix and converts it to seconds
+            Example:
+
+            The received output of systemd-analzye time is the following:
+
+            Startup finished in 505ms (userspace)
+            graphical.target reached after 453ms in userspace
+
+            For getting the startup time, use prefix = "Startup finished in " and the function will return 0.505s
+            """
+            i = stdout.find(prefix)
+            if i < 0:
+                raise ValueError(f"prefix {prefix} not found")
+            t = (
+                stdout[i + len(prefix) + 1 :].strip().split(" ")[0]
+            )  # pick the the next word after the prefix, i.e. this should contain the time
+            if t.endswith("ms"):
+                return float(t[:-2]) / 1000.0
+            elif t.endswith("s"):
+                return float(t[:-1])
+            raise ValueError("time unit not recognized")
+
+        time = auto_container.connection.run_expect(
+            [0], "systemd-analyze time"
+        )
+        startup = extract_time(time.stdout, "Startup finished in ")
+        assert startup <= STARTUP_LIMIT, "Startup threshold exceeded"
+        target = extract_time(time.stdout, ".target reached after ")
+        assert (
+            target <= TARGET_LIMIT
+        ), "Reaching systemd target threshold exceeded"
+
+    def test_systemd_nofailed_units(self, auto_container):
+        """
+        Ensure there are no failed systemd units
+        """
+        output = auto_container.connection.run_expect(
+            [0], "systemctl list-units --state=failed"
+        )
+        assert (
+            "0 loaded units listed" in output.stdout
+        ), "failed systemd units detected"
+
+    def test_systemd_detect_virt(
+        self, auto_container, host, container_runtime
+    ):
+        """
+        Ensure :command:`systemd-detect-virt` detects the container runtime
+        """
+        output = auto_container.connection.run_expect(
+            [0], "systemd-detect-virt"
+        ).stdout
+        runtime = container_runtime.runner_binary
+        assert (
+            runtime in output
+        ), f"systemd-detect-virt failed to detect {runtime}"
+
+    def test_journald(self, auto_container):
+        """
+        Ensure :command:`journald` works correctly
+        """
+
+        # Check that we reached at least the multiuser target
+        journal = auto_container.connection.run_expect(
+            [0], "journalctl --boot"
+        )
+        assert (
+            "Reached target Multi-User System" in journal.stdout
+        ), "Multi-User target was not reached"
+
+    def test_hostnamectl(self, auto_container, host, container_runtime):
+        """
+        Ensure :command:`hostnamectl` works correctly by asserting expected values
+        """
+
+        hostnamectl = auto_container.connection.run_expect([0], "hostnamectl")
+        # Process the printed values to a string map
+        values = TestSystemd._split_values(hostnamectl.stdout)
+        assert values["Chassis"] == "container", "Chassis mismatch"
+        assert (
+            "SUSE Linux Enterprise Server" in values["Operating System"]
+        ), "Missing SUSE tag in Operating system"
+        assert (
+            values["Virtualization"] == container_runtime.runner_binary
+        ), "Virtualization tag mismatch"
+
+    def test_timedatectl(self, auto_container):
+        """
+        Ensure :command:`timedatectl` works as expected and the container timezone is UTC
+        """
+        output = auto_container.connection.run_expect(
+            [0], "timedatectl"
+        ).stdout
+        assert "Time zone: UTC" in output, "Time zone not set to UTC"
+
+        # Check that the reported timestamp for UTC and local time match the system time
+        def check_timestamp(pattern, timestamp, delta):
+            """Checks the timedatectl output for the given pattern against the given timestamp
+            e.g. use the "Universal time" as pattern and datetime.utcnow() to check for the UTC time
+            """
+            grep = [line for line in output.split("\n") if pattern in line]
+            assert (
+                len(grep) == 1
+            ), f"{pattern} not present in timedatectl output"
+            tsp = (
+                grep[0].strip()[len(pattern) + 2 :].strip()
+            )  # Extract actual timestamp
+            tsp = datetime.datetime.strptime(tsp, "%a %Y-%m-%d %H:%M:%S UTC")
+            assert (
+                abs(tsp - timestamp) < delta
+            ), f"timedatectl diff exceeded for {pattern}"
+
+        check_timestamp(
+            "Universal time",
+            datetime.datetime.utcnow(),
+            datetime.timedelta(seconds=59),
+        )
+        # In the container the Local time is expected to be UTC
+        check_timestamp(
+            "Local time",
+            datetime.datetime.utcnow(),
+            datetime.timedelta(seconds=59),
+        )
+
+    def test_no_loginctl_sessions(self, auto_container):
+        """
+        Ensure :command:`loginctl` contains no logins
+        """
+        loginctl = auto_container.connection.run_expect([0], "loginctl")
+        assert (
+            "No sessions" in loginctl.stdout
+        ), "Assert no sessions are present failed"
+
+    @staticmethod
+    def _split_values(output: str, sep: str = ":") -> Dict[str, str]:
+        """Auxilliary function to process the given output into a key:value map, one entry by line"""
+        ret = {}
+        for line in output.split("\n"):
+            tmp = line.strip().split(sep)
+            if len(tmp) == 2:
+                name, value = tmp[0].strip(), tmp[1].strip()
+                ret[name] = value
+        return ret


tests/test_init.py
```diff
@@ -26,3 +26,208 @@ def test_systemd_present(container):
...
+
+@pytest.mark.parametrize("container", [INIT_CONTAINER], indirect=True)
+@pytest.mark.skipif(
+    DOCKER_SELECTED
+    and (int(LOCALHOST.package("systemd").version.split(".")[0]) >= 248),
+    reason="Running systemd in docker is broken as of systemd 248, see https://github.com/moby/moby/issues/42275",
+)
+def test_systemd_nofailed_units(container):
+    """
+    Check that there are not failed systemd units present
```
dcermak, line 127:
```suggestion
    Check that there are no failed systemd units present
```

tests/test_init.py
```diff
@@ -26,3 +26,208 @@ def test_systemd_present(container):
...
+)
+def test_systemd_detect_virt(container, host, container_runtime):
+    """
+    Check that systemd-detect-virt correctly detects the used container runtime
+    """
+    output = container.connection.run_expect([0], "systemd-detect-virt")
+    runtime = container_runtime.runner_binary
+    assert (
+        runtime in output.stdout
+    ), f"systemd-detect-virt failed to detect {runtime}"
```
dcermak, line 151:
```suggestion
    output = container.connection.run_expect([0], "systemd-detect-virt").stdout
    runtime = container_runtime.runner_binary
    assert (
        runtime in output
    ), "systemd-detect-virt failed to detect {runtime}"
```

tests/test_init.py
```diff
@@ -26,3 +26,208 @@ def test_systemd_present(container):
...
+    and (int(LOCALHOST.package("systemd").version.split(".")[0]) >= 248),
+    reason="Running systemd in docker is broken as of systemd 248, see https://github.com/moby/moby/issues/42275",
+)
+def test_hostnamectl(container, host, container_runtime):
+    """
+    Check that hostnamectl prints the expected values
+    """
+
+    # Process the given output into a key:value map, one entry by line
+    def split_values(output, sep=":"):
```
dcermak, line 184:
```suggestion
    def split_values(output, sep=":"):
        """Process the given output into a key:value map, one entry by line"""
```

tests/test_init.py
```diff
@@ -26,3 +26,208 @@ def test_systemd_present(container):
...
+        ret = {}
+        for line in output.split("\n"):
+            line = line.strip()
+            i = line.find(sep)
+            if i < 0:
+                continue
+            name, value = line[:i].strip(), line[i + 1 :].strip()
+            ret[name] = value
```
dcermak, line 192: How about something like this?
```suggestion
            tmp = line.strip().split(sep)
            if len(tmp) != 2:
                continue
            name, value = tmp
            ret[name] = value
```

tests/test_init.py
```diff
@@ -26,3 +26,208 @@ def test_systemd_present(container):
...
+        return ret
+
+    hostnamectl = container.connection.run_expect([0], "hostnamectl")
+    # Process the printed values to a string map
+    values = split_values(hostnamectl.stdout)
+    assert values["Chassis"] == "container", f"Chassis mismatch"
+    assert (
+        "SUSE Linux Enterprise Server" in values["Operating System"]
+    ), f"Missing SUSE tag in Operating system"
```
dcermak, line 201:
```suggestion
    ), "Missing SUSE tag in Operating system"
```

tests/test_init.py
```diff
@@ -26,3 +26,208 @@ def test_systemd_present(container):
...
+            if i < 0:
+                continue
+            name, value = line[:i].strip(), line[i + 1 :].strip()
+            ret[name] = value
+        return ret
+
+    hostnamectl = container.connection.run_expect([0], "hostnamectl")
+    # Process the printed values to a string map
+    values = split_values(hostnamectl.stdout)
+    assert values["Chassis"] == "container", f"Chassis mismatch"
```
dcermak, line 198:
```suggestion
    assert values["Chassis"] == "container", "Chassis mismatch"
```

tests/test_init.py
```diff
@@ -26,3 +26,208 @@ def test_systemd_present(container):
...
+    assert (
+        "SUSE Linux Enterprise Server" in values["Operating System"]
+    ), f"Missing SUSE tag in Operating system"
+    assert (
+        values["Virtualization"] == container_runtime.runner_binary
+    ), f"Virtualization tag mismatch"
```
dcermak, line 204:
```suggestion
    ), "Virtualization tag mismatch"
```

tests/test_init.py
```diff
@@ -26,3 +26,208 @@ def test_systemd_present(container):
...
+@pytest.mark.skipif(
+    DOCKER_SELECTED
+    and (int(LOCALHOST.package("systemd").version.split(".")[0]) >= 248),
+    reason="Running systemd in docker is broken as of systemd 248, see https://github.com/moby/moby/issues/42275",
+)
+def test_timedatectl(container):
+    """
+    Check that timedatectl can be invoked without issues
+    """
+    container.connection.run_expect([0], "timedatectl")
```
dcermak, line 217: Maybe add a check that the time & date are roughly correct?

tests/test_init.py
```diff
@@ -26,3 +26,208 @@ def test_systemd_present(container):
...
+
+
+@pytest.mark.parametrize("container", [INIT_CONTAINER], indirect=True)
+@pytest.mark.skipif(
+    DOCKER_SELECTED
+    and (int(LOCALHOST.package("systemd").version.split(".")[0]) >= 248),
+    reason="Running systemd in docker is broken as of systemd 248, see https://github.com/moby/moby/issues/42275",
+)
+def test_loginctl(container):
```
dcermak, line 226:
```suggestion
def test_no_loginctl_sessions(container):
```

tests/test_init.py
```diff
@@ -26,3 +26,208 @@ def test_systemd_present(container):
...
+def test_journald(container):
+    """
+    Check that journald behaves correctly
+    """
+
+    # Check that we reached at least the multiuser target
+    journal = container.connection.run_expect([0], "journalctl --boot -e")
+    assert (
+        "Reached target Multi-User System" in journal.stdout
+    ), f"Multi-User target was not reached"
```
dcermak, line 169:
```suggestion
    ), "Multi-User target was not reached"
```

tests/test_init.py
```diff
@@ -26,3 +26,208 @@ def test_systemd_present(container):
...
+    and (int(LOCALHOST.package("systemd").version.split(".")[0]) >= 248),
+    reason="Running systemd in docker is broken as of systemd 248, see https://github.com/moby/moby/issues/42275",
+)
+def test_journald(container):
+    """
+    Check that journald behaves correctly
+    """
+
+    # Check that we reached at least the multiuser target
+    journal = container.connection.run_expect([0], "journalctl --boot -e")
```
dcermak, line 166: I think you can omit the `-e`, as that only makes a difference when run in interactive mode

tests/test_init.py
```diff
@@ -26,3 +26,208 @@ def test_systemd_present(container):
...
+def test_systemd_nofailed_units(container):
+    """
+    Check that there are not failed systemd units present
+    """
+    output = container.connection.run_expect(
+        [0], "systemctl list-units --state=failed"
+    )
+    assert (
+        "0 loaded units listed" in output.stdout
+    ), "failed systemd units detected"
```
dcermak, line 116: Isn't this the same as the function in line 119?

tests/test_init.py
```diff
@@ -26,3 +26,208 @@ def test_systemd_present(container):
...
+    DOCKER_SELECTED
+    and (int(LOCALHOST.package("systemd").version.split(".")[0]) >= 248),
+    reason="Running systemd in docker is broken as of systemd 248, see https://github.com/moby/moby/issues/42275",
+)
+def test_systemd_boottime(container):
+    """Check that the startup time of the container is reasonable"""
+
+    # internal helper function to extract the time from systemd-analyze time
+    # extracts a timestamp with a given prefix and converts it to seconds
+    def extract_time(stdout, prefix):
```
dcermak, line 42: I would also suggest to include the expected form of the output of systemd-analyze and explain how and what you are extracting.
```suggestion
    def extract_time(stdout, prefix):
        """internal helper function to extract the time from systemd-analyze time
        extracts a timestamp with a given prefix and converts it to seconds"""
```

tests/test_init.py
```diff
@@ -26,3 +26,208 @@ def test_systemd_present(container):
...
+        i = stdout.find(prefix)
+        if i < 0:
+            raise ValueError("prefix not found")
```
dcermak, line 45:
```suggestion
            raise ValueError(f"prefix '{prefix}' not found")
```

tests/test_init.py
```diff
@@ -26,3 +26,208 @@ def test_systemd_present(container):
...
+            raise ValueError("time unit not recognized")
+
+    time = container.connection.run_expect([0], "systemd-analyze time")
+    startup = extract_time(time.stdout, "Startup finished in ")
+    assert startup < 5, "Startup threshold exceeded"
+    target = extract_time(time.stdout, ".target reached after ")
+    assert target < 5, "Reaching systemd target threshold exceeded"
+
+
+@pytest.mark.parametrize("container", [INIT_CONTAINER], indirect=True)
```
dcermak, line 65: Is this a duplicate of the first function that you added?

dcermak (changes requested):
Looks pretty good, I've added a few suggestions.

I would additionally suggest that you use the [`pytestmark`](https://docs.pytest.org/en/6.2.x/reference.html#globalvar-pytestmark) global variable to just skip this module with docker & systemd >= 248. and Please use the `auto_container` fixture instead of `container`. There's really no need to repeat the parametrize for every single test function :wink:

tests/test_init.py
grisu48: Done

tests/test_init.py
grisu48: Yep, `output` suggests that this is the better handling.

tests/test_init.py
grisu48: I added an output example and described how this function is expected to behave. It is better now, but still reads a bit bulky to me, if you have a better suggestion feel free to comment.

tests/test_init.py
grisu48: good idea. Added it now.

tests/test_init.py
```diff
@@ -3,26 +3,196 @@
...
+        """Check that :command:`systemctl` is in ``$PATH``, that :command:`systemctl
+        status` works and that :file:`/etc/machine-id` exists.
 
-    """
-    assert container.connection.exists("systemctl")
-    assert container.connection.file("/etc/machine-id").exists
-    assert container.connection.run_expect([0], "systemctl status")
+        This test is currently broken due to `moby/moby#42275
+        <https://github.com/moby/moby/issues/42275>`_ with :command:`docker` and
+        :command:`systemd` >= 248.
```
dcermak, line 30: This should be moved into the test class docstring.

tests/test_init.py
```diff
@@ -3,26 +3,196 @@
...
+                grep[0].strip()[len(pattern) + 2 :].strip()
+            )  # Extract actual timestamp
+            tsp = datetime.datetime.strptime(tsp, "%a %Y-%m-%d %H:%M:%S UTC")
+            print(
+                "%s - %s"
+                % (
+                    tsp.strftime("%m/%d/%Y, %H:%M:%S"),
+                    timestamp.strftime("%m/%d/%Y, %H:%M:%S"),
+                )
+            )
```
dcermak, line 164: left over print statement
```suggestion
```

tests/test_init.py
```diff
@@ -3,26 +3,196 @@
...
+            # Ensure the time difference is the higher - lower
+            diff = (tsp - timestamp) if tsp > timestamp else (timestamp - tsp)
```
dcermak, line 166: Did you just reinvent `math.abs()` :wink:?

tests/test_init.py
```diff
@@ -3,26 +3,196 @@
...
+        """
+        Check that :command:`loginctl` contains no logins
+        """
+        loginctl = auto_container.connection.run_expect([0], "loginctl")
+        assert (
+            "No sessions" in loginctl.stdout
+        ), "Assert no sessions are present failed"
+
+    @classmethod
+    def _split_values(cls, output, sep=":"):
```
dcermak, line 190: This needn't be a class method
```suggestion
    @staticmethod
    def _split_values(output, sep=":"):
```

dcermak (commented):
Looks good, just a few nitpicks here and there

dcermak:
> I don't know why the pipeline is complaining about the code not being `black` formatted, IMHO it is:
> 
> ```
> ./format.sh --check
> All done! ✨ 🍰 ✨
> 29 files would be left unchanged.
> ```

purge your `.tox` directly and run `tox -e format`

tests/test_init.py
```diff
@@ -3,26 +3,196 @@
...
+    def test_systemd_present(self, auto_container):
+        """Check that :command:`systemctl` is in ``$PATH``, that :command:`systemctl
+        status` works and that :file:`/etc/machine-id` exists.
 
-    """
-    assert container.connection.exists("systemctl")
-    assert container.connection.file("/etc/machine-id").exists
-    assert container.connection.run_expect([0], "systemctl status")
+        This test is currently broken due to `moby/moby#42275
+        <https://github.com/moby/moby/issues/42275>`_ with :command:`docker` and
```
jlausuch, line 29: why this issue again if it's already mentioned in line 17?

tests/test_init.py
grisu48: I wish it would have been that simple. Without this I sometimes got weird constructs like day `-1` which makes the following comparison fail.

tests/test_init.py
grisu48: Should not be there.

tests/test_init.py
grisu48: I removed it as it is already present in the pymark

tests/test_init.py
grisu48: Nono, I meant it should not be there, I will remove it and you will see it being gone once I git pushed it.

tests/test_init.py
```diff
@@ -3,26 +3,196 @@
...
+    def test_systemd_boottime(self, auto_container):
+        """Ensure the container startup time is reasonable"""
+
+        STARTUP_LIMIT = 5  # Container startup time limit in seconds
+        TARGET_LIMIT = (
+            5  # Limit for the systemd target to be reached in seconds
+        )
+
+        def extract_time(stdout, prefix):
+            """internal helper function to extract the time from systemd-analyze time
```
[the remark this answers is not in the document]
grisu48: I guess not but since it's an internal function also I'm wondering if that would be even desired.

The reason I left this within the `test_systemd_boottime` function is that this is a helper function that is useful exclusively in the scope of this function.

tests/test_init.py
m-dati: couldn't it be as by Dan suggestion ? (import math):
` diff = math.abs(tsp - timestamp)`

grisu48: No because it's an instance of `datetime.Timedelta` and not a float.

grisu48: Example:

```python
import datetime
import time

t1 = datetime.datetime.now()
time.sleep(2)
t2 = datetime.datetime.now()
print(t1-t2) # -1 day, 23:59:58.999138
print(t2-t1) # 0:00:01.000862

d1 = t1-t2
d2 = t2-t1

delta = datetime.timedelta(seconds=1)
print(d1 < delta) # True
print(d2 < delta) # False
```

tests/test_init.py
```diff
@@ -2,27 +2,184 @@
...
+    def test_hostnamectl(self, auto_container, host, container_runtime):
+        """
+        Check expected values from :command:`hostnamectl`
+        """
+
+        hostnamectl = auto_container.connection.run_expect([0], "hostnamectl")
+        # Process the printed values to a string map
+        values = TestSystemd._split_values(hostnamectl.stdout)
+        assert values["Chassis"] == "container", "Chassis mismatch"
+        assert (
```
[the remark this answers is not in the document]
jlausuch: ok, I have just tested `registry.opensuse.org/devel/bci/sle-15-sp3/images/bci/bci-init:latest` and it also contains ` Operating System: SUSE Linux Enterprise Server 15 SP3`, which makes sense, as they are based on sle.

tests/test_init.py
```diff
@@ -2,27 +2,184 @@
...
+            "SUSE Linux Enterprise Server" in values["Operating System"]
+        ), "Missing SUSE tag in Operating system"
+        assert (
+            values["Virtualization"] == container_runtime.runner_binary
```
jlausuch, line 127: in the previous test you are doing:
```
        assert (
            runtime in output
        ), f"systemd-detect-virt failed to detect {runtime}"
```
so, using `runtime` only. Why `container_runtime.runner_binary` is needed here?

tests/test_init.py
m-dati: yes, anyway abs is Supported Operation od datetime:https://docs.python.org/3/library/datetime.html#datetime.timedelta

dcermak: @grisu48 jfyi as @m-dati correctly noted: `abs(d1) == abs(d2)` is True

tests/test_init.py
dcermak: Because of this:
```ShellSession
❯ podman exec -it gifted_mcclintock hostnamectl|grep Virtualization
    Virtualization: podman
```

tests/test_init.py
grisu48: Oh nice, this works.

I tried it first with `math.abs` and that does not work, however `abs` does indeed what it should be doing.

tests/test_init.py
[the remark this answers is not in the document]
grisu48: I guess this means your comment has been addressed?

dcermak:
@grisu48 please add `tox -e format --check` to your pre-push git hooks :joy:

dcermak (changes requested):
One assertion doesn't look right to me, otherwise this is ready to be merged :muscle:

tests/test_init.py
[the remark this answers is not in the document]
jlausuch: Yes, but I'm not sure if we should make it more generic and just match `suse` word only instead of full string with `enterprise...`. 
I don't know how future BCI containers will look like in OBS, if they will be "openSUSE" based or what... @dcermak do you know?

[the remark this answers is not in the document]
jlausuch: Ok, fair enough. I'm just too cautious :) 
Let's fix it when it breaks something (if it does).

tests/test_init.py
```diff
@@ -2,27 +2,183 @@
...
+            grep = [line for line in output.split("\n") if pattern in line]
+            assert (
+                len(grep) == 1
+            ), f"{pattern} not present in timedatectl output"
+            tsp = (
+                grep[0].strip()[len(pattern) + 2 :].strip()
+            )  # Extract actual timestamp
+            tsp = datetime.datetime.strptime(tsp, "%a %Y-%m-%d %H:%M:%S UTC")
+            # Ensure the time difference is the higher - lower
+            assert abs(tsp - timestamp), f"timedatectl diff exceeded for {pattern}"
```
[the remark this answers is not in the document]
grisu48: Note to myself: Don't commit before your second cup of coffee :facepalm:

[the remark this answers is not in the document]
grisu48: Fixed now.

jlausuch (approved):
Good job :)

dcermak (changes requested):
Please run `tox -e format`

grisu48:
Did that while you have been complaining ;-) Should be good now, after the pipeline passes

dcermak:
@jlausuch @grisu48 the last pipeline (https://github.com/SUSE/BCI-tests/runs/6875884458?check_suite_focus=true) failed with:
```
E       AssertionError: Unexpected exit code 1 for CommandResult(command=b'systemd-analyze time', exit_status=1, stdout=None, stderr=b"\x1b[0;1;31mBootup is not yet finished (org.freedesktop.systemd1.Manager.FinishTimestampMonotonic=0).\x1b[0m\n\x1b[0;1;31mPlease try again later.\x1b[0m\n\x1b[0;1;31mHint: Use 'systemctl list-jobs' to see active jobs\x1b[0m\n")
```
This could result in a flaky test, but maybe we should add a `HEALTHCHECK` to init instead. WDYT?

grisu48:
I will have a look at this. Let me have another look, I would like to not push tests with flanky results. Marking as draft until resolved.

grisu48:
https://github.com/SUSE/BCI-dockerfile-generator/pull/9 should indeed prevent race conditions where the system target has not yet been reached. Thanks for the improvement.

I've pushed the black formatted test run and think we can proceed with this PR.

dcermak (changes requested):
Could it be that you force pushed some of your improvements away? There's a bunch of print statements back in there

grisu48:
> Could it be that you force pushed some of your improvements away? There's a bunch of print statements back in there

Oh damn, the rebase did something weird. Part of the latest changes got removed. Let me check, something went horribly wrong here.

Happy Monday btw :coffee:

grisu48:
Fixed. Please review again and sorry for the fuzz @dcermak

tests/test_init.py
```diff
@@ -2,27 +2,184 @@
...
+        """
+        Ensure :command:`loginctl` contains no logins
+        """
+        loginctl = auto_container.connection.run_expect([0], "loginctl")
+        assert (
+            "No sessions" in loginctl.stdout
+        ), "Assert no sessions are present failed"
+
+    @staticmethod
+    def _split_values(output: str, sep: str = ":") -> dict[str, str]:
```
dcermak, line 177: Using `dict` for a dictionary type is not supported with older python versions.
```suggestion
    def _split_values(output: str, sep: str = ":") -> Dict[str, str]:
```

dcermak (changes requested):
You have to use `Dict` and not `dict`.

tests/test_init.py
grisu48: Done.

tests/test_init.py
```diff
@@ -2,27 +2,185 @@
...
+        assert auto_container.connection.exists("systemctl")
+        assert auto_container.connection.file("/etc/machine-id").exists
+        assert auto_container.connection.run_expect([0], "systemctl status")
 
-    """
-    assert container.connection.exists("systemctl")
-    assert container.connection.file("/etc/machine-id").exists
-    assert container.connection.run_expect([0], "systemctl status")
+    def test_systemd_boottime(self, auto_container):
+        """Ensure the container startup time is below defines limits"""
```
dcermak, line 35: Typo
```suggestion
        """Ensure the container startup time is below 5 seconds"""
```

dcermak (approved):
Thank you!

PR #143, opened 2022-06-16
Add fips testsuite

dcermak:


tests/test_fips.py
```diff
@@ -0,0 +1,215 @@
...
+COPY --from=builder /usr/lib64/libssl.so.1.1 /usr/lib64/
+COPY --from=builder /lib64/libz.so.1 /usr/lib64/
+RUN /bin/fips-test sha256
+"""
+
+#: Simple C program that will calculate the hash of a fixed string using openSSL.
+#: The hash to be used is passed as the first parameter to the binary.
+#: This program has been taken from
+#: `<https://www.openssl.org/docs/manmaster/man3/EVP_DigestInit.html>`_
+FIPS_TEST_DOT_C = r"""#include <openssl/evp.h>
```
grisu48, line 62: Is there no way of putting this in an external file? Working on inline code is somehow cumbersome.

grisu48 (commented):
One small comment, otherwise LGTM!

tests/test_fips.py
dcermak: There is, it's now under `tests/files/`

PR #149, opened 2022-07-04
Only add .Net containers to metadata tests on x86_64

dcermak:
The .Net containers only get build for x86_64, thus we cannot add them to the
metadata tests as they otherwise cause the assertion:
```
assert len(ALL_CONTAINERS) == len(
    IMAGES_AND_NAMES
), "IMAGES_AND_NAMES must have all containers from BASE_CONTAINERS"
```
to fail.


jlausuch (approved):
LGTM. Thanks for the quick fix.

PR #151, opened 2022-07-05
BCI Java: files and directories test

ilausuch:
https://progress.opensuse.org/issues/110689
VR: http://marvin5.arch.suse.cz/tests/702

### DIFF
--- a/tests/test_openjdk.py
+++ b/tests/test_openjdk.py
@@ -74,6 +74,10 @@ def test_jdk_version(container, java_version):
             ["I am the thread 1", "I am the thread 2"],
         ),
         ("time", ["All OK"]),
+        (
+            "files_and_directories",
+            ["All OK"],
+        ),
     ],
 )
 def test_jdk_extended(
@@ -87,6 +91,7 @@ def test_jdk_extended(
     It covers:
     - threading tests
     - java time and date tests
+    - files and dirs tests
     The validation is done checking the exit code (0) and checking that some
     expected strings can be found on the stdout of the execution.
     """
--- a/tests/trainers/java/files_and_directories.java
+++ b/tests/trainers/java/files_and_directories.java
@@ -1,0 +1,101 @@
+/**
+Test to check the Java File module.
+It consists on:
+- Text and binary file read and write
+- File operations (e.g. exists, permissions, rename and delete)
+- Directory operations (e.g. read content, creation)
+*/
+import java.io.*;
+import java.io.IOException;
+import java.util.UUID;
+
+class Test {
+    static String textContent = "Content of the text file";
+
+    static void createTextFile(String filepath) throws IOException{
+        FileWriter file = new FileWriter(filepath);
+        file.write(Test.textContent);
+        file.close();
+    }
+
+    static void readTextFile(String filepath) throws IOException {
+        FileReader file = new FileReader(filepath);
+        char [] a = new char[1024];
+        int length = file.read(a);
+        String output = "";
+        for(int i=0; i<length; i++)
+            output = output + a[i];
+        file.close();
+    }
+
+    static void createBinaryFile(String inputFile, String outputFile) throws IOException {
+        InputStream inputStream = new FileInputStream(inputFile);
+        OutputStream outputStream = new FileOutputStream(outputFile);
+
+        int byteRead = -1;
+        while ((byteRead = inputStream.read()) != -1) {
+            outputStream.write((byte)byteRead);
+        }
+
+        inputStream.close();
+        outputStream.close();
+    }
+
+    static void findFile(String dirpath, String filename) throws Exception {
+        boolean exists = false;
+        File[] files = new File(dirpath).listFiles();
+        for (int i = 0; i < files.length; i++) {
+            if (files[i].getName().equals(filename))
+                exists = true;
+        }
+        
+        assert exists;
+    }
+
+    public static void main(String[] args) {
+        UUID fileuuid=UUID.randomUUID();
+        
+        String dirpath = "/tmp";
+        String textfile =  dirpath + "/" + fileuuid + ".txt";
+        String binaryfile = dirpath + "/" + fileuuid + ".bin";
+        String binaryfile2 = dirpath + "/" + fileuuid + ".data";
+        
+
+        try {
+            // Text files tests
+            Test.createTextFile(textfile);
+            Test.readTextFile(textfile);
+
+            // Binary files tests
+            Test.createBinaryFile(textfile, binaryfile);
+            Test.readTextFile(binaryfile);
+
+            // Files tests
+            assert new File(textfile).length() == Test.textContent.length();
+            assert new File(textfile).exists() == true;
+            assert new File(textfile).isFile();
+            new File(binaryfile).renameTo(new File(binaryfile2));
+            new File(binaryfile2).delete();
+            assert new File(textfile).canRead() == true;
+            assert new File(textfile).canWrite() == true;
+            new File(textfile).setExecutable(true);
+            assert new File(textfile).canExecute() == true;
+
+            // Directories tests
+            assert new File(dirpath).isDirectory();
+            new File(dirpath + "/extra").mkdir();
+            Test.findFile(dirpath, fileuuid + ".txt");
+            
+            // General
+            new File(dirpath).getFreeSpace();
+            File f = File.createTempFile("" + fileuuid, ".tmp");
+
+            System.out.println("All OK");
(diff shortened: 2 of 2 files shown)


dcermak:
Please rebase your branch on top of main and format the source code.

tests/trainers/java/files_and_directories.java
```diff
@@ -0,0 +1,185 @@
...
+        } catch (IOException ex) {
+            return "KO - " + ex.getMessage();
+        }
+    }
+
+    private String readTextFile(String filepath) {
+        try {
+            FileReader file = new FileReader(filepath);
+            char [] a = new char[50];
+            file.read(a);
```
grisu48, line 29: `read` returns the number of read characters, you can use this to avoid having to trim the output:

```suggestion
            final int n = file.read(a);
            final String output = new String(a, 0, n);
```

Also I would increase the size of the char array to something like 1024 - the malus is almost negligible, but you avoid future issues that arise from the buffer size being really really small.

tests/trainers/java/files_and_directories.java
```diff
@@ -0,0 +1,185 @@
...
+            file.close();
+            return "OK";
+        } catch (IOException ex) {
+            return "KO - " + ex.getMessage();
+        }
+    }
+
+    private String readTextFile(String filepath) {
+        try {
+            FileReader file = new FileReader(filepath);
```
grisu48, line 27: I would move the FileReader outside, so that you can use a `finally` block

```suggestion
        FileReader file = null;
        try {
            FileReader file = new FileReader(filepath);
            // snip
        } catch (IOException ex) {
            // snip
        } finally {
            if (file != null) {
                file.close();
            }
        }
```

Otherwise the file handle would not be closed in case e.g. an `IOException` is raised while doing the read.

grisu48: In this case this is only a style question, so feel free to also ignore this comment if you feel it's not applicable.

tests/trainers/java/files_and_directories.java
```diff
@@ -0,0 +1,185 @@
...
+                return "OK";
+            else
+                return "KO - file content doesn't match with '" + this.textContent + "'";
+        } catch (IOException ex) {
+            return "KO - " + ex.getMessage();
+        }
+    }
+
+    private String createBinaryFile(String inputFile, String outputFile) {
+        try (
```
grisu48, line 46: You're missing the `close` for the `inputStream` and the `outputStream` in the whole block.

grisu48 (commented):
Nice coverage increase for the java tests! I've added some suggestions that might give you an idea how you can still add some improvements.

tests/trainers/java/files_and_directories.java
ilausuch: done, thanks!

tests/test_openjdk.py
```diff
@@ -71,7 +71,11 @@ def test_jdk_version(container, java_version):
...
             ["I am the thread 1", "I am the thread 2"],
-        )
+        ),
+        (
+            "files_and_directories",
+            ["ALL OK"],
+        ),
     ],
 )
 def test_jdk_extended(
```
dcermak, line 81: Please update the docstring of this function.

tests/trainers/java/files_and_directories.java
```diff
@@ -0,0 +1,216 @@
...
+            }catch(IOException ex2) {}
+            try {
+                outputStream.close();
+            }catch(IOException ex2) {}
+            
+        }
+    }
+
+    private String length(String filepath, int minSize){
+        return new File(filepath).length() == minSize ? "OK" : "KO - file " + filepath + " should have " + minSize + " bytes";
```
dcermak, line 92: Wouldn't it be better to use a test framework for this? You're really just re-implementing `assert` here.

dcermak (changes requested):
The python documentation needs adjustment. When it comes to the java code I have no real idea if this is good style or not, so I cannot really judge that.

tests/test_openjdk.py
ilausuch: done

tests/trainers/java/files_and_directories.java
```diff
@@ -0,0 +1,170 @@
...
+        int byteRead = -1;
+        while ((byteRead = inputStream.read()) != -1) {
+            outputStream.write((byte)byteRead);
+        }
+        
+        inputStream.close();
+        outputStream.close();
+    }
+
+    private void length(String filepath, int minSize) throws IOException, Exception {
```
dcermak, line 47: Isn't `IOException` a subclass of `Exception`?

tests/trainers/java/files_and_directories.java
```diff
@@ -0,0 +1,170 @@
...
+        if (new File(filepath).length() != minSize)
+            throw new Exception ("File " + filepath + " should have " + minSize + " bytes");
+    }
+
+    private void exists(String filepath, boolean expected) throws IOException, Exception {
```
dcermak, line 52: Wouldn't it make sense to declare all these methods as static? They don't access any class variables, or do they?

dcermak (commented):
LGTM overall, I've just added a few comments (which might be dumb, because I don't know java)

tests/trainers/java/files_and_directories.java
ilausuch: I am adding assert when it's possible

tests/trainers/java/files_and_directories.java
ilausuch: Ok, I agree, let's simplify the code

tests/trainers/java/files_and_directories.java
ilausuch: yes, could be static

tests/trainers/java/files_and_directories.java
```diff
@@ -0,0 +1,170 @@
...
+        for(int i=0; i<length; i++)
+            output = output + a[i];
+        file.close();
+    }
+
+    private void createBinaryFile(String inputFile, String outputFile) throws IOException {
+        InputStream inputStream = null;
+        OutputStream outputStream = null;
+        inputStream = new FileInputStream(inputFile);
+        outputStream = new FileOutputStream(outputFile);
```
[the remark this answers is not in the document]
ilausuch: because previously was inside a try...catch Now fixed

ilausuch:
I implemented all the suggestions

tests/trainers/java/files_and_directories.java
```diff
@@ -0,0 +1,156 @@
...
+        new File(inputFile).setExecutable(executable);
+    }
+
+    static void isDirectory(String dirpath) throws Exception {
+        assert new File(dirpath).isDirectory();
+    }
+
+    static void mkdir(String dirpath) throws IOException {
+        new File(dirpath).mkdir();
+    }
```
dcermak, line 91: I know I'm now deep in the nitpick territory, but since these are now trivial one-liners, don't you think we can just include them like they are in `main`?

grisu48: Since java is bloaty or verbosy, I think Ivan is following the recommended way here :wink: 

I'd leave it like this, as the overall structure of his java class is to separate tests in individual methods. Neither syntactical nor structural there is a malus for doing so, and it appears to be closer to the approach that e.g. junit has (each test in it's own method) but without using junit per se.

To me this is bloaty, but well structured, which IMHO more valuable than saving a couple of verbose method signatures.

ilausuch: Thanks both for your comments and share your point of view. The initial idea was to split that on different methods, but it was before the first review when each method consisted on more than one operation. So in this point I think inline is a good option

tests/trainers/java/files_and_directories.java
```diff
@@ -0,0 +1,101 @@
...
+        try {
+            // Text files tests
+            Test.createTextFile(textfile);
+            Test.readTextFile(textfile);
+
+            // Binary files tests
+            Test.createBinaryFile(textfile, binaryfile);
+            Test.readTextFile(binaryfile);
+
+            // Filtes tests
```
dcermak, line 73: I guess that's a typo?
```suggestion
            // Files tests
```

ilausuch: yes thanks, done

dcermak (changes requested):
LGTM, but you now have merge conflicts with `main`. Please rebase your branch.

dcermak (approved):
LGTM, thank you!

PR #159, opened 2022-07-14
Add simple smoke tests for the Rust containers

dcermak:


grisu48 (dismissed):
Good as always nothing to add. I especially like the version test that checks if the rust version in the container matches the expected version.

LGTM Dan, nice work! :rocket:

jlausuch (approved):

PR #160, opened 2022-07-15
Increase max size for BCI base for aarch64 and ppc64le

jlausuch:
After trying to enable bci tests for base image, I found the following errors:
aarch64:
```
>       assert (
            container_runtime.get_image_size(auto_container.image_url_or_id)
            < BASE_CONTAINER_MAX_SIZE[LOCALHOST.system_info.arch] * 1024 * 1024
        )
E       AssertionError: assert 137972672.0 < ((130 * 1024) * 1024)
```

s390x:
```
>       assert (
            container_runtime.get_image_size(auto_container.image_url_or_id)
            < BASE_CONTAINER_MAX_SIZE[LOCALHOST.system_info.arch] * 1024 * 1024
        )
E       AssertionError: assert 160373180.0 < ((150 * 1024) * 1024)
```

### DIFF
--- a/tests/test_base.py
+++ b/tests/test_base.py
@@ -21,8 +21,8 @@ from bci_tester.runtime_choice import DOCKER_SELECTED
 #: size limits of the base container per arch in MiB
 BASE_CONTAINER_MAX_SIZE: Dict[str, int] = {
     "x86_64": 120,
-    "aarch64": 130,
-    "ppc64le": 150,
+    "aarch64": 135,
+    "ppc64le": 160,
     "s390x": 120,
 }
 


jlausuch:
I might need to increase also ppc...

grisu48:
Are we sure that there is not something wrong with the images if they suddenly increased in disk size?

jlausuch:
> Are we sure that there is not something wrong with the images if they suddenly increased in disk size?

I am not 100% sure, but we never ran BCI tests for these architectures, and these values seem obsoleted... I kind of expected this, similar to what happened to other images, e.g. https://gitlab.suse.de/dancermak/bci-tests/-/merge_requests/58

We have already some size monitoring in place, so I guess it's safe to merge this and start counting a potential size increase from this point of time where we have green tests and a monitor tool.

grisu48 (commented):
Ok, then you got my post-merge approval.

PR #169, opened 2022-09-14
BCI java test with cassandra

ilausuch:
Improve the BCI tests using a real application as cassandra DB.

https://progress.opensuse.org/issues/115043
VR: http://openqa.suse.de/tests/9622447

### DIFF
--- a/tests/test_openjdk.py
+++ b/tests/test_openjdk.py
@@ -1,9 +1,13 @@
 """Tests of the OpenJDK base container."""
+import os
+import re
+import time
 from dataclasses import dataclass
 from dataclasses import field
 
 import pytest
 from pytest_container import DerivedContainer
+from pytest_container import Version
 from pytest_container.container import container_from_pytest_param
 
 from bci_tester.data import OPENJDK_11_CONTAINER
@@ -15,6 +19,10 @@ HOST_TEST_DIR = "tests/trainers/java/"
 DOCKERF_EXTENDED = f"""
 WORKDIR {CONTAINER_TEST_DIR}
 COPY {HOST_TEST_DIR} {CONTAINER_TEST_DIR}
+"""
+
+DOCKERF_CASSANDRA = """
+RUN zypper in -y tar gzip awk git
 """
 
 CONTAINER_IMAGES = [
@@ -32,6 +40,18 @@ CONTAINER_IMAGES_EXTENDED = [
         id=container.id,
     )
     for container in CONTAINER_IMAGES
+]
+
+CONTAINER_IMAGES_CASSANDRA = [
+    pytest.param(
+        DerivedContainer(
+            base=container_from_pytest_param(container),
+            containerfile=DOCKERF_CASSANDRA,
+        ),
+        marks=container.marks,
+        id=container.id,
+    )
+    for container in [OPENJDK_11_CONTAINER]
 ]
 
 
@@ -152,3 +172,68 @@ def test_jdk_extended(
 
     for check in params.expected_err_strings:
         assert check in testout.stderr
+
+
+@pytest.mark.parametrize(
+    "container_per_test",
+    CONTAINER_IMAGES_CASSANDRA,
+    indirect=["container_per_test"],
+)
+def test_jdk_cassandra(container_per_test):
+    """Starts the Cassandra DB and executes some write and read tests
+    using the cassandra-stress
+    """
+
+    logs = "/var/log/cassandra.log"
+
+    cassandra_versions = container_per_test.connection.check_output(
+        "git ls-remote --tags https://gitbox.apache.org/repos/asf/cassandra.git"
+    )
+
+    cassandra_version = Version(0, 0, 0)
+    for line in cassandra_versions.splitlines():
+        match = re.search(r"cassandra-(\d).(\d).(\d)$", line)
+        if match:
+            cur_ver = Version(
+                int(match.group(1)), int(match.group(2)), int(match.group(3))
+            )
+            if cur_ver > cassandra_version:
+                cassandra_version = cur_ver
+
+    container_per_test.connection.run_expect(
+        [0],
+        f"curl -OL https://downloads.apache.org/cassandra/{cassandra_version}/apache-cassandra-{cassandra_version}-bin.tar.gz",
+    )
+
+    container_per_test.connection.run_expect(
+        [0],
+        f"tar xzvf apache-cassandra-{cassandra_version}-bin.tar.gz >/dev/null",
+    )
+
+    container_per_test.connection.run_expect(
+        [0],
+        f"export JAVA_HOME=/usr/ && cd apache-cassandra-{cassandra_version}/ && bin/cassandra -R | tee {logs}",
+    )
+
+    check_str = "state jump to NORMAL"
+    found = False
+    for t in range(80):
+        time.sleep(10)
+        if (
+            check_str
+            in container_per_test.connection.file(logs).content_string
+        ):
+            found = True
+            break
+
+    assert found, f"{check_str} not found in {logs}"
+
+    container_per_test.connection.run_expect(
+        [0],
+        f"export JAVA_HOME=/usr/ && cd apache-cassandra-{cassandra_version}/tools/bin/ && ./cassandra-stress write n=1",
+    )
+
+    container_per_test.connection.run_expect(
+        [0],
+        f"export JAVA_HOME=/usr/ && cd apache-cassandra-{cassandra_version}/tools/bin/ && ./cassandra-stress read n=1",
+    )


tests/test_openjdk.py
```diff
@@ -156,3 +173,48 @@ def test_jdk_extended(
...
+def test_jdk_cassandra(
+    container_per_test,
+    host,
+    container_runtime: OciRuntimeBase,
+):
+    """TODO info
+    """
+
+    container_per_test.connection.run_expect(
+        [0], "curl -OL https://downloads.apache.org/cassandra/4.0.6/apache-cassandra-4.0.6-bin.tar.gz"
```
dcermak, line 192: Is it a good idea to hardcode the version here?

ilausuch: I did in a different way

tests/test_openjdk.py
```diff
@@ -19,6 +20,10 @@
 COPY {host_test_dir} {container_test_dir}
 """
 
+DOCKERF_CASSANDRA = f"""
```
mimi1vx, line 23: why f-string here ?

ilausuch: yes, not need

ilausuch: fixed

tests/test_openjdk.py
```diff
@@ -156,3 +174,65 @@ def test_jdk_extended(
...
         assert check in testout.stderr
+
+
+@pytest.mark.parametrize(
+    "container_per_test",
+    CONTAINER_IMAGES_CASSANDRA,
+    indirect=["container_per_test"],
+)
+@pytest.mark.parametrize(
+    "cassandra_version", [(os.getenv("APPLICATION_VERSION", "4.0.6"))]
```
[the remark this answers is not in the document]
ilausuch: Ok, thanks. The perl part will be in an other PR (in https://github.com/os-autoinst/os-autoinst-distri-opensuse) So, this PR is complete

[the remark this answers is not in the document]
m-dati: yes, thank you, LGTM

tests/test_openjdk.py
```diff
@@ -156,3 +174,65 @@ def test_jdk_extended(
...
+)
+def test_jdk_cassandra(
+    container_per_test,
+    cassandra_version,
+    host,
+    container_runtime: OciRuntimeBase,
```
dcermak, line 191: Unused parameters
```suggestion
```

tests/test_openjdk.py
```diff
@@ -156,3 +174,65 @@ def test_jdk_extended(
...
+        ):
+            found = True
+            break
+
+    assert (
+        "state jump to NORMAL"
+        in container_per_test.connection.check_output(
+            "cat /var/log/cassandra.log"
+        )
+    )
```
dcermak, line 228: Why this duplication and not just `assert found`?

tests/test_openjdk.py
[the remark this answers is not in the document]
dcermak: This is a super generic environment variable name, please consider something different, e.g. `CASSANDRA_VERSION`.

Also, I don't particularly like this approach to be honest, as it just adds complexity for the user. Imho the program should try to get the latest version and not offload this onto the user. You could e.g. use this: `git ls-remote --tags https://gitbox.apache.org/repos/asf/cassandra.git` to get the tags from the cassandra repo and use the latest non-beta/alpha tag (or even include those).

dcermak (changes requested):
There's a few minor issues left.

tests/test_openjdk.py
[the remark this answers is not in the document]
ilausuch: Thanks that helped a lot

tests/test_openjdk.py
ilausuch: Done

tests/test_openjdk.py
ilausuch: Done

tests/test_openjdk.py
```diff
@@ -156,3 +174,56 @@ def test_jdk_extended(
...
+    CONTAINER_IMAGES_CASSANDRA,
+    indirect=["container_per_test"],
+)
+def test_jdk_cassandra(container_per_test):
+    """Starts the Cassandra DB and executes some write and read tests
+    using the cassandra-stress
+    """
+
+    cassandra_version = container_per_test.connection.check_output(
+        'git ls-remote --tags https://gitbox.apache.org/repos/asf/cassandra.git | grep -v alpha | grep -v beta | grep -v "{}" | tail -n 1 | cut -d "-" -f2'
```
dcermak, line 190: nitpick: if it is simpler to perform the parsing in python, then I'd suggest to use that instead, I find that usually much more obvious.
I would nevertheless suggest to add a simple sanity check, e.g. that you get only one version and/or that it matches a simple regex or something along those lines. It will make debugging of issues like `curl: failed to fetch https://downloads.apache.org/cassandra/remote hung up/apache-cassandra-remote hung up-bin.tar.gz 404` much easier :wink:

dcermak (changes requested):
The CI fails in the lint and openjdk stages. These need to be green, before we can merge.

ilausuch:
The VR is green. This could be merged if no more reviews

dcermak:
> The VR is green. This could be merged if no more reviews

github actions are not, please fix them.

tests/test_openjdk.py
```diff
@@ -156,3 +175,71 @@ def test_jdk_extended(
...
+        x = re.search("cassandra-(\d).(\d).(\d)$", line)
+        if x:
+            order_value = (
+                int(x.group(1)) * 1000 + int(x.group(2))
+            ) * 1000 + int(x.group(3))
+            if order_value > cassandra_version_order_value:
+                cassandra_version = (
+                    x.group(1) + "." + x.group(2) + "." + x.group(3)
+                )
+                cassandra_version_order_value = order_value
```
dcermak, line 207: How about using [`pytest_container.Version`](https://github.com/dcermak/pytest_container/blob/e67f174d9e1546a721ccd7e1f7e74c46b1579bf7/src/pytest_container/runtime.py#L64) for this?

You will not have to perform the parsing yourself then and you get the version comparison for "free".

dcermak:
@ilausuch Please rebase on main before you push the next change, that should fix the `lint` job :crossed_fingers:

tests/test_openjdk.py
ilausuch: done

tests/test_openjdk.py
```diff
@@ -152,3 +172,70 @@ def test_jdk_extended(
...
+        x = re.search(r"cassandra-(\d).(\d).(\d)$", line)
+        if x:
+            if Version(
+                int(x.group(1)), int(x.group(2)), int(x.group(3))
+            ) > Version(int(version[0]), int(version[1]), int(version[2])):
+                version[0] = x.group(1)
+                version[1] = x.group(2)
+                version[2] = x.group(3)
+
+    cassandra_version = ".".join(version)
```
dcermak, line 203: I think something like this should work and is much simpler:
```suggestion
    cassandra_version = Version(0, 0, 0)
    for line in cassandra_versions.splitlines():
        cur_ver = Version.parse(line.replace("cassandra-", ""))
        if cur_ver > cassandra_version:
            cassandra_version = cur_ver
```

ilausuch: Done with some changes in your code.
`cur_ver = Version.parse(line.replace("cassandra-", ""))` doesn't work because some special cases. e.g.
`ValueError: Invalid version string: 66ee8b58577a6b61c1195eaa797d14ce38042ea7	refs/tags/0.3.0-final`

dcermak:
@ilausuch LGTM, but the CI still fails with:
```
>       assert found
E       assert False
```
PTAL

tests/test_openjdk.py
dcermak: Ah, yeah, makes sense, I did not actually try my code code :sweat_smile:

dcermak (changes requested):
Please rebase on main

ilausuch:
> @ilausuch LGTM, but the CI still fails with:
> 
> ```
> >       assert found
> E       assert False
> ```
> 
> PTAL

It has to be that it needs more time on the CI machines. I testing to increase the delay for timeout

tests/test_openjdk.py
```diff
@@ -152,3 +172,66 @@ def test_jdk_extended(
...
+        if (
+            "state jump to NORMAL"
+            in container_per_test.connection.check_output(
+                "cat /var/log/cassandra.log"
+            )
+        ):
+            found = True
+            break
+
+    assert found
```
dcermak, line 227: You might want to include the log file in the assertion message to give you a better error message.

tests/test_openjdk.py
```diff
@@ -152,3 +172,66 @@ def test_jdk_extended(
...
+
+    found = False
+    for t in range(40):
+        time.sleep(10)
+        if (
+            "state jump to NORMAL"
+            in container_per_test.connection.check_output(
+                "cat /var/log/cassandra.log"
+            )
+        ):
```
dcermak, line 223: nitpick: https://testinfra.readthedocs.io/en/latest/modules.html?highlight=file#testinfra.modules.file.File.contains is a better fit for this

dcermak:
@ilausuch rebasing on main will resolve the openjdk-devel test failures

ilausuch:
I am doing and anaylisis and trying to reproduce the errors. One interesting case is docker + 15.4 (Only for this case)

```
Unrecognized VM option 'UseConcMarkSweepGC'\nError: Could not create the Java Virtual Machine.\nError: A fatal exception has occurred. Program will exit
```

dcermak:
> I am doing and anaylisis and trying to reproduce the errors. One interesting case is 15.4 with java 17
> 
> ```
> Unrecognized VM option 'UseConcMarkSweepGC'\nError: Could not create the Java Virtual Machine.\nError: A fatal exception has occurred. Program will exit
> ```

The concurrent mark & sweep garbage collector has been removed in JDK 15 or 16, so it will not be available in java 17. And we have no Java 17 in 15.3, so that's why this error appears only for SP4

tests/test_openjdk.py
ilausuch: done

tests/test_openjdk.py
ilausuch: done

ilausuch:
This is ready

dcermak (approved):
LGTM, thanks!

PR #179, opened 2022-10-24
Add test for L3 support

grisu48:
Modify the metadata test to check for techpreview and L3 support for matching containers.

Related ticket: https://progress.opensuse.org/issues/119191


bci_tester/data.py
```diff
@@ -393,6 +412,38 @@ def create_BCI(
     BUSYBOX_CONTAINER,
 ]
 
+## Metadata labels according to https://confluence.suse.com/display/ENGCTNRSTORY/SLE+BCI+Image+Overview
+# Containers that are in tech preview
```
dcermak, line 416:
```suggestion
#: Metadata labels according to https://confluence.suse.com/display/ENGCTNRSTORY/SLE+BCI+Image+Overview
#: Containers that are in tech preview
```

bci_tester/data.py
```diff
@@ -393,6 +412,38 @@ def create_BCI(
...
+    DOTNET_ASPNET_3_1_CONTAINER,
+    DOTNET_ASPNET_6_0_CONTAINER,
+    DOTNET_SDK_3_1_CONTAINER,
+    DOTNET_SDK_6_0_CONTAINER,
+    DOTNET_RUNTIME_3_1_CONTAINER,
+    DOTNET_RUNTIME_6_0_CONTAINER,
+    NODEJS_12_CONTAINER,
+]
+
+# Containers with L3 support
```
dcermak, line 427:
```suggestion
#: Containers with L3 support
```

dcermak (changes requested):
Just a few nitpicks

bci_tester/data.py
```diff
@@ -289,13 +289,32 @@ def create_BCI(
...
-    create_BCI(
-        build_tag=f"bci/rust:{rust_version}",
-        image_type="dockerfile",
-        available_versions=["15.4"],
-    )
-    for rust_version in ("1.59", "1.60", "1.61", "1.62")
+    RUST_1_59_CONTAINER,
+    RUST_1_60_CONTAINER,
+    RUST_1_61_CONTAINER,
+    RUST_1_62_CONTAINER,
```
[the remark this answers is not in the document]
grisu48: We need to address specific containers for the labels afterwards, that's why I un-looped them.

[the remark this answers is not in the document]
grisu48: Wait let me be more specific: Currently we don't need them but once Rust 1.63 lands we will need them. I prepared the variables already for this case.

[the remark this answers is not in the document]
dcermak: You can do this in python:
```python
RUST_1_59_CONTAINER, RUST_1_60_CONTAINER, RUST_1_61_CONTAINER, RUST_1_62_CONTAINER = RUST_CONTAINERS
```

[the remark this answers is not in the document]
grisu48: Done.

bci_tester/data.py
```diff
@@ -393,6 +399,38 @@ def create_BCI(
     BUSYBOX_CONTAINER,
 ]
 
+#: Metadata labels according to https://confluence.suse.com/display/ENGCTNRSTORY/SLE+BCI+Image+Overview
+#: Containers that are in tech preview
+TECHPREVIEW_CONTAINERS = [
```
dcermak, line 404: Rust is missing.

I would suggest to add an assertion to the metadata test that `len(TECHPREVIEW_CONTAINERS) + len(L3_CONTAINERS) == len(ALL_CONTAINERS)`

grisu48: Dammit, yes. I've added the rust containers by version manually because 1.63 will be in L3.

That assertion will not work, because e.g. golang 1.16 and 1.17 are not in techpreview nor in i3, at least not according to the reference sheet in confluence.

grisu48:
The test failures in metadata are expected, because the label of the already released containers does not match the expectation of the new labels.

bci_tester/data.py
dcermak: Everything that is not in l3 is techpreview. unsupported just means eol in that table, but the label will still be techpreview.

grisu48: I changed the PR such, that it now checks all containers in `L3_CONTAINERS` for the l3 label, all other containers need to have the `techpreview` label. This is simpler, does not require additional checks to ensure that we are not missing containers and should work as well.

bci_tester/data.py
```diff
@@ -393,6 +399,28 @@ def create_BCI(
     BUSYBOX_CONTAINER,
 ]
 
+#: Metadata labels according to https://confluence.suse.com/display/ENGCTNRSTORY/SLE+BCI+Image+Overview
+#: Containers that are not in L3 are supposed to be in techpreview
```
dcermak, line 403: I don't think that this comment makes sense here.
```suggestion
```

tests/test_metadata.py
```diff
@@ -373,21 +374,38 @@ def test_image_type_label(
...
 )
-def test_supportlevel_label(
+def test_techpreview_label(
     container_data: ParameterSet,
 ):
-    """Check that all containers (except for the base container) have the label
-    ``com.suse.supportlevel`` set to ``true``.
-
+    """Check that containers in non-l3 container have the label
+    ``com.suse.supportlevel`` set to ``techpreview``.
```
dcermak, line 383:
```suggestion
    """Check that containers that are not L3 supported have the label
    ``com.suse.supportlevel`` set to ``techpreview``.
```

tests/test_metadata.py
```diff
@@ -373,21 +374,38 @@ def test_image_type_label(
...
 
+@pytest.mark.parametrize(
+    "container_data",
+    [cont for cont in L3_CONTAINERS],
+)
+def test_l3_label(
+    container_data: ParameterSet,
+):
+    """Check that containers in L3 the label
+    ``com.suse.supportlevel`` set to ``l3``.
```
dcermak, line 400:
```suggestion
    """Check that containers under L3 support have the label
    ``com.suse.supportlevel`` set to ``l3``.
```

dcermak (commented):
LGTM, I've added a few suggestions for the docs

dcermak:
> The test failures in metadata are expected, because the label of the already released containers does not match the expectation of the new labels.

This will break your openQA tests though.

dcermak:
@grisu48 Please rebase on main as well

grisu48:
> This will break your openQA tests though.

They are already broken and need sthis change, e.g. https://openqa.suse.de/tests/overview?distri=sle&version=15-SP4&build=18.6_golang-1.19-image&groupid=444

dcermak (changes requested):
The metadata tests are failing on 15.3.

grisu48:
> The metadata tests are failing on 15.3.

Those failures are expected, since the released images are not yet having the new expected label.

grisu48:
Thanks Dan!

PR #186, opened 2022-11-23
Stop testing language pack containers on anything older than 15.4

dirkmueller:


dcermak:
Why do you want to disable Ruby 2.5 only and not all other language stack containers that are SP4 only?

dcermak (changes requested):
`available_versions` has been introduced to take care of containers that are not being built for all sle versions. If you don't want to run the tests on 15.3, simply don't run them.

dirkmueller:
@dcermak the github actions here are simply doing a test matrix over all tools, all container runtimes and all OS versions. skipping those that do not make sense hence seemed sensible to do. 

so you're suggesting to split the github action into base + 15.3-only jobs and then run the rest only for 15.4?

or are you suggesting to default to 15.4?

dcermak:
> @dcermak the github actions here are simply doing a test matrix over all tools, all container runtimes and all OS versions. skipping those that do not make sense hence seemed sensible to do.

Ok, but this is the wrong approach for that.

> so you're suggesting to split the github action into base + 15.3-only jobs and then run the rest only for 15.4?

That won't work easily given how we setup the CI. If you really want to exclude certain tests, I'd suggest to exclude them in the github action: https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#jobsjob_idstrategymatrixexclude

> or are you suggesting to default to 15.4? like e.g. #188

That's a good idea, but unrelated to this.

dirkmueller:
It's just one case. I would also prefer to not build the containers that we don't deliver in OBS, which means it needs to be excluded from testing also outside GitHub actions.

I'll submit a new PR for the github actions update and keep this around for later.

dirkmueller:
not needed anymore after #191. thanks @dcermak !

PR #191, opened 2022-11-30
Test language & application stack containers only on latest os version

dcermak:
supersedes #186

### DIFF
--- a/bci_tester/data.py
+++ b/bci_tester/data.py
@@ -1,4 +1,5 @@
 #!/usr/bin/env python3
+import enum
 import os
 from datetime import timedelta
 from typing import List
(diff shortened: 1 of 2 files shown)


bci_tester/data.py
```diff
@@ -28,6 +29,14 @@
 #: places
 OS_VERSION = os.getenv("OS_VERSION", "15.4")
```
dirkmueller, line 31: add ciomment that the last entry is the special and hence the list needs to be sorted, or sort it explicitly..

bci_tester/data.py
```diff
@@ -28,6 +29,14 @@
...
 
+OS_VERSIONS = ["15.3", "15.4"]
```
dirkmueller, line 32:
```suggestion
OS_VERSIONS = ("15.3", "15.4")
```

bci_tester/data.py
```diff
@@ -28,6 +29,14 @@
...
+
+if OS_VERSION not in OS_VERSIONS:
```
dirkmueller, line 34: maybe we should rename `OS_VERSIONS` to `ALLOWED_OS_VERSIONS` ?

dirkmueller (approved):

PR #192, opened 2022-11-30
Switch to Rust 1.64 & 1.65, add .Net 7 containers 

dcermak:


bci_tester/data.py
```diff
@@ -288,20 +300,37 @@ def create_BCI(
...
 
 RUST_CONTAINERS = [
     create_BCI(
         build_tag=f"bci/rust:{rust_version}",
         image_type="dockerfile",
         available_versions=["15.4"],
     )
-    for rust_version in ("1.59", "1.60", "1.61", "1.62")
+    for rust_version in (
+        "1.59",
```
jlausuch, line 317: do we want to have the old versions too?

jlausuch (dismissed):

PR #193, opened 2022-12-07
Test the environment variable JAVA_HOME for openjdk & openjdk-devel

dcermak:
This is a regression test for https://bugzilla.suse.com/show_bug.cgi?id=1206128

These tests will fail before https://github.com/SUSE/BCI-dockerfile-generator/pull/156 is merged and deployed.

### DIFF
--- a/tests/test_openjdk.py
+++ b/tests/test_openjdk.py
@@ -1,6 +1,7 @@
 """Tests of the OpenJDK base container."""
-import os
+import os.path
 import re
+import stat
 import time
 from dataclasses import dataclass
 from dataclasses import field
@@ -9,6 +10,7 @@ import pytest
 from pytest_container import DerivedContainer
 from pytest_container import Version
 from pytest_container.container import container_from_pytest_param
+from pytest_container.container import ContainerData
 from pytest_container.runtime import LOCALHOST
 
 from bci_tester.data import OPENJDK_11_CONTAINER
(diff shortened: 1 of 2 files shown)


dirkmueller:
rebased and applied black reformatting to fix CI failure.

tests/test_openjdk.py
```diff
@@ -82,6 +84,51 @@ def test_jdk_version(container, java_version):
     )
 
 
+def test_java_home(auto_container: ContainerData):
+    """Check that the environment variable ``JAVA_HOME`` is set to the correct
+    value.
+
```
grisu48, line 90:
```suggestion
Regression test following https://bugzilla.suse.com/show_bug.cgi?id=1206128
```

I would add the bsc for reference.

tests/test_openjdk.py
```diff
@@ -82,6 +84,51 @@ def test_jdk_version(container, java_version):
...
+    java_bindir_path = get_env_var("JAVA_BINDIR")
+    java_bindir = auto_container.connection.file(java_bindir_path)
+    assert java_bindir.exists and java_bindir.is_directory
+
+    java = auto_container.connection.file(
+        os.path.join(java_bindir_path, "java")
+    )
+    assert java.exists and java.is_file and java.mode & stat.S_IXUSR
+
```
grisu48, line 117: Since we are here already, I would even execute the binary once, e.g.

```suggestion
auto_container.connection.run_expect([0], java_bindir_path + "/java --version")
```

WDYT?

grisu48 (commented):
LGTM! I've added two suggestions

dirkmueller (approved):

PR #194, opened 2022-12-07
Switch to SLE BCI language pack for unit testing

dirkmueller:
The GitHub runners no longer support 3.6, and 3.11 is the default python version nowadays.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -48,18 +48,20 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        python_version: ["3.6", "3.7", "3.8", "3.9", "3.10"]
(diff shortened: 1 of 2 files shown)


dirkmueller:
@dcermak why do we need to run the tests-unittests on 3.6? that makes no sense to me

dirkmueller:
then we'd have to run the unit tests inside a sle bci 3.6 container (which makes little sense for the sle bci testsuite?!)

we ran out of policy: https://github.com/actions/runner-images#support-policy . only 5 python versions are supported, and 3.11 was added.

dirkmueller:
I'm testing an approach to build the unit tests in a container..

dcermak:
@dirkmueller please rebase on main

PR #195, opened 2022-12-07
Improve podman setup for build in latest ubuntu-latest

dirkmueller:
The podman builds are now failing because the registries.d directory no longer exists. install podman explicitly and also create the directory for more reliable builds.


dcermak (approved):
Thanks!

PR #196, opened 2022-12-09
Add dotnet containers to L3

grisu48:
All dotnet containers should container the L3 label. Add them  to the L3_CONTAINER definition list.


dcermak (approved):

grisu48:
Metadata failures are expected, because the released containers don't have the new label yet.

dcermak (changes requested):
Blocking this until we have the final call from PM

grisu48:
Following the feedback from Michal this PR is obsolete.

PR #197, opened 2022-12-13
Improve cassandra version extraction logic

dirkmueller:
The regular expression wasn't formatted properly ('.' needs backslashes to match the literal dot rather than any character)

Plus run curl with -f to avoid downloading 404 urls.

### DIFF
--- a/tests/test_openjdk.py
+++ b/tests/test_openjdk.py
@@ -247,7 +247,7 @@ def test_jdk_cassandra(container_per_test):
 
     cassandra_version = Version(0, 0, 0)
     for line in cassandra_versions.splitlines():
-        match = re.search(r"cassandra-(\d).(\d).(\d)$", line)
+        match = re.search(r"cassandra-(\d+)\.(\d+)\.(\d+)$", line)
         if match:
             cur_ver = Version(
                 int(match.group(1)), int(match.group(2)), int(match.group(3))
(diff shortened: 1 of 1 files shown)


tests/test_openjdk.py
```diff
@@ -245,19 +245,23 @@ def test_jdk_cassandra(container_per_test):
...
             cur_ver = Version(
                 int(match.group(1)), int(match.group(2)), int(match.group(3))
             )
-            if cur_ver > cassandra_version:
-                cassandra_version = cur_ver
+            if cur_ver > best_ver:
+                best_ver = cur_ver
+                cassandra_version = (
+                    f"{match.group(1)}.{match.group(2)}.{match.group(3)}"
+                )
```
dcermak, line 260: This variable is pointless, you can as well use `best_ver` at this point. FYI, `Version` has a `__str__()` that converts it to exactly what you need here.

dcermak: That's a bug in the `Version` class and not intended at all. I will push a fix upstream.

dcermak: https://github.com/dcermak/pytest_container/pull/117

dcermak: fix has been merged

dirkmueller: The documentation needs updating: `./src/pytest_container/runtime.py`

```
    >>> Version(1, 0) > Version(0, 1)
    True
    >>> Version(1, 0) == Version(1, 0, 0)
    True
```

dcermak: For comparisons and equality checks, I have opted to set the patch level to 0 if it is `None`. From a SemVer perspective `1.0` and `1.0.0` should be equal.

Changing the code so that `Version(1, 0) != Version(1, 0, 0)` would have the unfortunate consequence that such versions cannot be ordered anymore: while they are not equal, neither of them is larger or smaller than the other…

I'm open to suggestions how to properly handle this

dirkmueller: I guess it is okay the way it is now. thanks for the fix!

dcermak (approved):

PR #198, opened 2022-12-16
Add test for distribution image

mloviska:
Add a simple test for [distribution-image](https://build.suse.de/package/view_file/SUSE:SLE-15-SP4:Update:BCI/distribution-image/distribution-image.kiwi?expand=1), that creates a custom container. The test container is pushed and pulled from the registry.

- ticket: [Create tests for the distribution-image](https://progress.opensuse.org/issues/120810)


tests/test_distribution.py
```diff
@@ -0,0 +1,55 @@
...
+    )
+    assert out == '{"repositories":["test_container"]}'
+
+    host.run_expect([0], f"{engine} rmi hello_registry")
+    host.run_expect(
+        [0],
+        f"{engine} pull --tls-verify=false localhost:{host_port}/test_container",
+    )
+    host.run_expect([0], f"{engine} run --rm test_container")
+    assert False
```
dcermak, line 55: :wink: 
```suggestion
```

tests/test_distribution.py
```diff
@@ -0,0 +1,55 @@
...
+    )
+    assert out == '{"repositories":[]}'
+
+    host.run(f"{engine} logs {auto_container_per_test.container_id}")
+    # build custom image to be pushed into the registry container
+    with open(tmp_path / "Containerfile", "w", encoding="utf-8") as cfile:
+        cfile.write(content)
+    host.run_expect(
+        [0],
+        f"cd {tmp_path} && {' '.join(container_runtime.build_command)} -t hello_registry",
```
dcermak, line 37: the build command fails with docker & podman

tests/test_distribution.py
mloviska: Oops, thanks! :D

dcermak:
With podman you're running into the issue that we have turned on image verification via https://github.com/SUSE/BCI-tests/blob/e9c8fe1842a2109b98ee1d155ff8705e5a6d6c2e/.github/workflows/ci.yaml#L129, but only included the key for `devel:BCI`. Could you pick a different image or reconfigure podman to also accept the opensuse signkey?

dirkmueller (approved):

dirkmueller:
This still needs https://github.com/dcermak/pytest_container/pull/120 to pass

mloviska:
> This still needs [dcermak/pytest_container#120](https://github.com/dcermak/pytest_container/pull/120) to pass

Oh, thanks for the update!

mloviska:
> With podman you're running into the issue that we have turned on image verification via
> 
> https://github.com/SUSE/BCI-tests/blob/e9c8fe1842a2109b98ee1d155ff8705e5a6d6c2e/.github/workflows/ci.yaml#L129
> , but only included the key for `devel:BCI`. Could you pick a different image or reconfigure podman to also accept the opensuse signkey?

Let's try out a devel image, and later on we can include opensuse signkey. Thanks for the hint

dcermak:
doesn't work with docker:
```
docker: Error response from daemon: invalid mode: U,noexec,Z.
```

mloviska:
> doesn't work with docker:
> 
> ```
> docker: Error response from daemon: invalid mode: U,noexec,Z.
> ```

Oh dear, IIUC docker is missing `U and noexec` flags for mounts, e.g. https://github.com/moby/moby/issues/7054, and it has not been implemented there for several years.

Is there any recommendation from SUSE towards our customers how they should use the container using `docker|podman`?

dcermak:
> Is there any recommendation from SUSE towards our customers how they should use the container using `docker|podman`?

podman is the recommended container runtime on ALP and SLES according to my knowledge.

dcermak:
> In such a case, would it be acceptable to run the test with `podman` only for now and re-thing what to do with `docker`? This way we can at least have some regular testing and release of the container.

That's ok from my PoV. Add a `skipIf` like here to the test function: https://github.com/SUSE/BCI-tests/blob/3cb5a194cefb3e080f2a00ecb06b2fdc31495892/tests/test_init.py#L18

tests/test_distribution.py
```diff
@@ -0,0 +1,56 @@
...
+from bci_tester.data import DISTRIBUTION_CONTAINER
+from bci_tester.runtime_choice import DOCKER_SELECTED
+
+CONTAINER_IMAGES = [DISTRIBUTION_CONTAINER]
+
+
+@pytest.mark.skipif(
+    DOCKER_SELECTED,
+    reason="docker does not support the same volume flags as podman (missing CHOWN and NOEXEC), e.g. https://github.com/moby/moby/issues/7054",
+)
```
dcermak, line 17: Given that the container image itself is unusable with docker, I would suggest to add this mark directly to the container.

mloviska: That makes perfect sense, done. Thanks for the suggestion.

dirkmueller:
> > podman is the recommended container runtime on ALP and SLES according to my knowledge.
> In such a case, would it be acceptable to run the test with `podman` only for now and re-thing what to do with `docker`? This way we can at least have some regular testing and release of the container.

No, we need to test with docker also, and there is no reason why that shouldn't work. In fact the main issue is the incorrect tagging in the test case, which breaks `docker push`. I'll post a fixup.

dirkmueller (dismissed):

tests/test_distribution.py
```diff
@@ -0,0 +1,56 @@
...
+    host_port = auto_container_per_test.forwarded_ports[0].host_port
+    content = """
+    FROM alpine:latest
+    CMD ["echo", "container from my local registry"]
+    """
+
+    out = host.check_output(
+        f"curl -sb -H 'Accept: application/json' http://localhost:{host_port}/v2/_catalog"
+    )
+    assert out == '{"repositories":[]}'
```
dcermak, line 28: recommendation: use the `json` module here, this test could easily break if they decide to format the json slightly differently

mloviska: Ok, will do. I guess same on L#52

tests/test_distribution.py
```diff
@@ -0,0 +1,56 @@
...
+
+    host.run(f"{engine} logs {auto_container_per_test.container_id}")
+    # build custom image to be pushed into the registry container
+    with open(tmp_path / "Containerfile", "w", encoding="utf-8") as cfile:
+        cfile.write(content)
+    host.run_expect(
+        [0],
+        f"""
+cd {tmp_path} && {' '.join(container_runtime.build_command)} \
+    -t {container_path} -f Containerfile .""",
```
dcermak, line 41: [textwrap.dedent](https://docs.python.org/3/library/textwrap.html#textwrap.dedent) is your friend for this ;-)

mloviska: Done, thanks for the suggestion

tests/test_distribution.py
mloviska: Updated

tests/test_distribution.py
```diff
@@ -22,10 +23,12 @@ def test_registry_service(
...
 
-    out = host.check_output(
-        f"curl -sb -H 'Accept: application/json' http://localhost:{host_port}/v2/_catalog"
+    out = json.loads(
+        host.check_output(
+            f"curl -sb -H 'Accept: application/json' http://localhost:{host_port}/v2/_catalog"
+        )
     )
-    assert out == '{"repositories":[]}'
+    assert len(out["repositories"]) == 0
```
dirkmueller, line 31: you're no longer checking whether there is no other output. the way to address the feedback from @dcermak without breaking existing test would have been something like

```python
assert str(out) == '{"repositories":[]}'
```

mloviska: The initial idea was to just check the content of the repository. The previous code was just a shortcut, and since there were no complaints it has survived there till now.

Now I am wondering whether we want to test the complete response from the registry, or we just value that it has started with empty volume and later on it contains only the image we have pushed there.

dcermak: I was thinking more of 
```suggestion
    assert out == {"repositories": []}
```

mloviska: Alright, I will update it accordingly

mloviska: I have managed to make it work with Dirk's suggestion. Thanks

dirkmueller (dismissed):

dirkmueller:
I've cleaned up the dedent stuff as well. now it looks good to me and should be ready for merging imho

bci_tester/data.py
```diff
@@ -411,6 +412,21 @@ def create_BCI(
...
+
+DISTRIBUTION_CONTAINER = create_BCI(
+    build_tag="suse/registry:2.8",
+    image_type="kiwi",
+    available_versions=["15.4"],
+    forwarded_ports=[PortForwarding(container_port=5000)],
+    default_entry_point=True,
+    volume_mounts=[
+        ContainerVolume(
+            host_path=f"docker-registry-{os.getpid()}",
```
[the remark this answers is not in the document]
dirkmueller: so how does one set a predictable name that persists for multiple tests then?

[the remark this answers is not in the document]
mloviska: I am working on this, I will push the changes after my meeting

dirkmueller (approved):

bci_tester/data.py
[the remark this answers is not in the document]
dirkmueller: already done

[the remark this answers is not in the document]
mloviska: > already done

Thanks! :)

dcermak (approved):
Thanks!

mloviska:
:partying_face: @dcermak, @dirkmueller thank you guys, your support and help was very much appreciated, you rock! :partying_face:

PR #200, opened 2023-01-11
Add rust 1.66 container

jlausuch:
https://progress.opensuse.org/issues/122956

### DIFF
--- a/bci_tester/data.py
+++ b/bci_tester/data.py
@@ -354,9 +354,13 @@ RUST_CONTAINERS = [
         image_type="dockerfile",
         available_versions=["15.4"],
     )
-    for rust_version in ("1.64", "1.65")
+    for rust_version in ("1.64", "1.65", "1.66")
 ]
-(RUST_1_64_CONTAINER, RUST_1_65_CONTAINER) = RUST_CONTAINERS
+(
+    RUST_1_64_CONTAINER,
+    RUST_1_65_CONTAINER,
+    RUST_1_66_CONTAINER,
+) = RUST_CONTAINERS
 
 INIT_CONTAINER = create_BCI(
     build_tag=f"bci/bci-init:{OS_VERSION}",
@@ -489,6 +493,7 @@ L3_CONTAINERS = [
     RUBY_25_CONTAINER,
     RUST_1_64_CONTAINER,
     RUST_1_65_CONTAINER,
+    RUST_1_66_CONTAINER,
     CONTAINER_389DS,
 ]
 
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -53,6 +53,7 @@ markers = [
     'suse/pcp_5.2.2',
     'bci/rust_1.64',
     'bci/rust_1.65',
+    'bci/rust_1.66',
     'suse/registry_2.8'
 ]
 


dcermak (changes requested):
This is not sufficient, you need to add the rust container to https://github.com/SUSE/BCI-tests/blob/bb59492ee6f7616245f37c07016c0313ac43d13a/bci_tester/data.py#L351 as well.

jlausuch:
> This is not sufficient, you need to add the rust container to
> 
> https://github.com/SUSE/BCI-tests/blob/bb59492ee6f7616245f37c07016c0313ac43d13a/bci_tester/data.py#L351
> as well.

Oops, I forgot. Thanks and updated :)

bci_tester/data.py
```diff
@@ -354,7 +354,7 @@ def create_BCI(
         image_type="dockerfile",
         available_versions=["15.4"],
     )
-    for rust_version in ("1.64", "1.65")
+    for rust_version in ("1.64", "1.65", "1.66")
 ]
 (RUST_1_64_CONTAINER, RUST_1_65_CONTAINER) = RUST_CONTAINERS
```
dcermak, line 359:
```suggestion
(RUST_1_64_CONTAINER, RUST_1_65_CONTAINER, RUST_1_66_CONTAINER) = RUST_CONTAINERS
```

And you need to add RUST_1_66_CONTAINER further below as well (and maybe to the metadata tests as well).

jlausuch: Amended. metadata uses the anchor `RUST_CONTAINERS`

bci_tester/data.py
```diff
@@ -489,6 +489,7 @@ def create_BCI(
     RUBY_25_CONTAINER,
     RUST_1_64_CONTAINER,
     RUST_1_65_CONTAINER,
+    RUST_1_66_CONTAINER
```
dcermak, line 492:
```suggestion
    RUST_1_66_CONTAINER,
```

dcermak:
And please change the commit message as well, it's no longer just adding the marker.

dcermak (approved):
LGTM, thanks!

jlausuch:
The failing pipeline says
```
AssertionError: Unexpected exit code 1 for CommandResult(command=b'docker pull registry.suse.com/bci/rust:1.66', exit_status=1, stdout=None, stderr=b'Error response from daemon: manifest for registry.suse.com/bci/rust:1.66 not found: manifest unknown: manifest unknown\n')
```
But this is expected, as the container is not tested yet, therefore not released. Shall we merge anyway?

dcermak:
> But this is expected, as the container is not tested yet, therefore not released. Shall we merge anyway?

It will also make your openQA tests fail. If you are ok with manually waving them until the container is published, then we can merge, otherwise add an xfail and we remove it once the container is published. Pick your poison :upside_down_face:

dirkmueller:
It feels like we need to merge this so that QA can approve the testing of rust1.66 and enable it? I think there is no harm in waiting

jlausuch:
> It feels like we need to merge this so that QA can approve the testing of rust1.66 and enable it? I think there is no harm in waiting

Yes, openQA should not care if the image is not in public registry as it will pull it from IBS.

dcermak:
> > It feels like we need to merge this so that QA can approve the testing of rust1.66 and enable it? I think there is no harm in waiting
> 
> Yes, openQA should not care if the image is not in public registry as it will pull it from IBS.

That's not the point of that metadata test. That test checks whether the reference is correct, that has nothing to do with where the image came from, but whether its canonical source is correct (which it can't be if the image has not been published yet).

PR #201, opened 2023-01-17
Remove containers that are not longer supported

jlausuch:
15-SP3 is in LTSS and only bci-base is supported, the rest are out of support.
Also, we only support the latest 2 version of rust and golang.


bci_tester/data.py
```diff
@@ -357,7 +345,6 @@ def create_BCI(
     for rust_version in ("1.64", "1.65", "1.66")
```
dcermak, line 345:
```suggestion
    for rust_version in ("1.65", "1.66")
```

jlausuch:
> The whole CI is broken
Please wait, working on it :)

bci_tester/data.py
```diff
@@ -203,13 +203,13 @@ def create_BCI(
 MINIMAL_CONTAINER = create_BCI(
     build_tag=f"bci/bci-minimal:{OS_VERSION}",
     image_type="kiwi",
-    available_versions=[OS_VERSION],
+    available_versions=["15.4"],
```
jlausuch, line 206: @dcermak can you help me figure out why the CI complains with
```
==================================== ERRORS ====================================
_____________________ ERROR collecting tests/test_build.py _____________________
'bci/bci-minimal_15.3' not found in `markers` configuration option
```
since I'm forcing only 1 version?

dcermak: You are creating a `bci/bci-minimal:15.3` and `bci/bci-minimal:15.4` as the build tag comes from the `OS_VERSION` environment variable.

dcermak:
@jlausuch Do I understand it correctly that you more or less want to drop 15.3 completely?

bci_tester/data.py
jlausuch: So, it should be hardcoded also here:
`build_tag=f"bci/bci-minimal:15.4",`

We probably need to change this once 15.5 is there and we need to support 15.3 (bci-base only), 15.4 (all) and 15.5 (all)

dcermak: No, just don't remove the build tag from pyproject.toml. with `available_versions=["15.4"]` that container will be automatically skipped for 15 SP3

jlausuch:
> 

not completely, but only bci-base:15.3

dcermak (approved):
LGTM now, let's see whether the CI agrees.

PR #202, opened 2023-02-10
Add platform1 testing

dirkmueller:


tests/test_all.py
```diff
@@ -52,7 +52,7 @@ def test_os_release(auto_container):
     """
     assert auto_container.connection.file("/etc/os-release").exists
 
-    for (var_name, value) in (
+    for var_name, value in (
```
dirkmueller, line 55: this is a `black` induced change, I can't get rid of it

PR #204, opened 2023-02-17
Enable BCI Base for 15-SP5 for Public Beta

jlausuch:


dcermak:
Just the base container and no others?

jlausuch:
> Just the base container and no others?

Others will come after GA if I'm not mistaken..

jlausuch:
> Just the base container and no others?

Ok, let me enable all the rest in one shot, anyway, we can enable/disable the openQA pipeline for those containers.

dcermak (changes requested):
This won't work for minimal, micro, busybox and init, they are set to be available for 15.4 only.

dcermak (changes requested):
init is missing

dcermak (changes requested):
```
tests/test_init.py:12: in <module>
    from bci_tester.data import INIT_CONTAINER
bci_tester/data.py:327: in <module>
    INIT_CONTAINER = create_BCI(
bci_tester/data.py:193: in create_BCI
    raise ValueError(
E   ValueError: Invalid os version for a language or application stack container: 15.5
```

dcermak (changes requested):
all the busybox tests are skipped due to the global pytestmark in test_busybox.py

PR #209, opened 2023-03-08
Remove the need for skopeo

dcermak:
pytest_container can now inspect containers directly. Thereby we do not need to rely on skopeo for the tests anymore

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -27,7 +27,6 @@ What do I need to contribute?
 * tox
 * go
 * docker and/or podman+buildah
-* skopeo (only required for the metadata tests)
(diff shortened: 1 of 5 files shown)


jlausuch (approved):
Nice! 1 dependency less.

dcermak:
Looks like the registry is currently not working :-/
Let's wait a bit and then restart the CI so that the tests are green again.

dcermak:
Looks like we had pcp tagged at version `5.2.2` and it just received a patch level bump to `5.2.5`. I have thus changed the version to `5`, so that we can avoid needlessly having to bump the version.

dcermak:
The metadata test failure is expected as the pcp container hasn't been released with this version yet

PR #211, opened 2023-03-10
Add postgresql tests

dcermak:
Depends on https://github.com/SUSE/BCI-dockerfile-generator/pull/278 being deployed. Before that the postgres tests will fail.


mloviska:
Shouldn't we try to use bind mount/volume with `postgresql` containers? Other than that, LGTM

dcermak:
> Shouldn't we try to use bind mount/volume with `postgresql` containers? Other than that, LGTM

That is already happening automagically. The postgres container has a `VOLUME /var/lib/postgresql/data` directive, so a volume will be created by podman/docker automatically.

mloviska (dismissed):

jlausuch (approved):
Neat! 
Do you think it would be worth adding more coverage like more crud operations with users, permissions, (besides tables and data)?

dcermak:
> Neat! Do you think it would be worth adding more coverage like more crud operations with users, permissions, (besides tables and data)?

No, I think it would be far more important to test that all the environment variables used in the docker-entrypoint do "the right thing^TM"

dcermak:
I have added some additional tests, please take a look. They will be failing currently though, because the necessary fixes are not yet deployed.

dirkmueller:
@dcermak needs rebase on top of the php tests

tests/test_postgres.py
```diff
@@ -0,0 +1,142 @@
...
+            (None, _postgres_user),
+            (None, _other_pg_user),
+            (POSTGRES_PASSWORD, _other_pg_pw),
+            (None, _pgdata),
+        ):
+            env = {"POSTGRES_PASSWORD": pw}
+            if pg_user:
+                env["POSTGRES_USER"] = pg_user
+            if pgdata:
+                env["PGDATA"] = pgdata
```
dirkmueller, line 59: why is there no "VOLUME {pgdata}" added to containerfile in this case? what is creating the volume here?

dcermak:
@dirkmueller could you help me review the postgres scenario test matrix? There's still plenty of failures but I lost track whether these are supported or unsupported scenarios…

dirkmueller (approved):

dirkmueller (approved):

PR #219, opened 2023-03-22
Run tests scheduled every other day

dirkmueller:
This way we can see sudden breakages outside PRs

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -2,6 +2,8 @@
 name: CI
 
 on:
+  schedule:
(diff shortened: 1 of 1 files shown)


dcermak:
I don't disagree, I'm just curious which breakage would have been caught by this in the past?

dirkmueller:
> I don't disagree, I'm just curious which breakage would have been caught by this in the past?

for example https://github.com/SUSE/BCI-tests/commit/016adec0cc3e2e1e639b6911677469d3f0fd859c but there were several.

PR #220, opened 2023-03-22
Run container tests in parallel

dirkmueller:


dcermak (dismissed):
I thought we were already doing that :see_no_evil:

dirkmueller:
> I thought we were already doing that see_no_evil

In a way we do that by running the test groups in parallel. the gain is not so much as the jobs take most of their time in setting up the ci job rather than running the tests. I think we need to group them more to have a better throughput. I'll play with that when I have some time.

dcermak:
> it does hang though if there are failures. which is not good :/

This looks like a bug in the locking of pytest_container, I'll take a look.

dcermak:
Should be fixed via https://github.com/dcermak/pytest_container/pull/135

dcermak:
Looks like the hang is gone

dcermak (approved):
This LGTM, is this ready to be merged @dirkmueller?

PR #222, opened 2023-03-22
Add openSUSE:Factory key

dcermak:
We need to add the signkey of openSUSE:Factory to the gpg keys as well, so that we can pull registry.opensuse.org/opensuse/nginx for the php tests. And for that, we need a newer podman…

Requires #221

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -236,7 +236,7 @@ jobs:
           EOF
 
       - name: Run tox job
-        run: sudo --preserve-env tox -e ${{ matrix.toxenv }} -- -n auto
+        run: sudo --preserve-env tox -e ${{ matrix.toxenv }} -- -n auto --reruns 3
         env:
           CONTAINER_RUNTIME: ${{ matrix.container_runtime }}
           OS_VERSION: ${{ matrix.os_version }}
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -156,9 +156,9 @@ jobs:
           EOF
 
           policy_json=$(cat /etc/containers/policy.json)
-          echo $policy_json | jq '.transports += { "docker": {"registry.opensuse.org": [{ "type": "signedBy", "keyType": "GPGKeys", "keyPath": "/tmp/devel_bci.key"}]}}' | sudo tee /etc/containers/policy.json
+          echo $policy_json | jq '.transports += { "docker": {"registry.opensuse.org": [{ "type": "signedBy", "keyType": "GPGKeys", "keyPaths": ["/etc/containers/devel_bci.key", "/etc/containers/opensuse_container.key"]}]}}' | sudo tee /etc/containers/policy.json
 
-          cat << EOF > /tmp/devel_bci.key
+          cat << EOF | sudo tee /etc/containers/devel_bci.key
           -----BEGIN PGP PUBLIC KEY BLOCK-----
 
           mQINBGQa2Y4BEAC+VBw/6hJCCd+JlrngmHvAS2dbzz0dk0dh6rK7mhuuQTmTbJex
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yaml
```diff
@@ -203,9 +214,43 @@ jobs:
...
+          Rq/lROEF5lx4SxB838JU4ezxD++BJXfBTE8JZmlGscXv74y9nCtSOZza8KOKj8ou
+          WRl739FMnx9jRd7HHj3TIyymoveODnZ7f3IElyyFsjBW3XuQ9XfpZrIkwHuaZV5M
+          6q2h+hgWNQ==
+          =nMh8
+          -----END PGP PUBLIC KEY BLOCK-----
+          EOF
 
       - name: Run tox job
-        run: sudo --preserve-env tox -e ${{ matrix.toxenv }} -- -n auto
+        run: sudo --preserve-env tox -e ${{ matrix.toxenv }} -- -n auto --reruns 3
```
dirkmueller, line 253: do we really want to do this? do we have such a flakey testsuite?

.github/workflows/ci.yaml
```diff
@@ -144,11 +144,22 @@ jobs:
...
         if: ${{ matrix.container_runtime == 'PODMAN' }}
         run: |
-          sudo apt update && sudo apt install podman
+          sudo mkdir -p /etc/apt/keyrings
+          curl -fsSL https://download.opensuse.org/repositories/devel:kubic:libcontainers:unstable/xUbuntu_$(lsb_release -rs)/Release.key \
+            | gpg --dearmor \
+            | sudo tee /etc/apt/keyrings/devel_kubic_libcontainers_unstable.gpg > /dev/null
+          echo \
+            "deb [arch=$(dpkg --print-architecture) signed-by=/etc/apt/keyrings/devel_kubic_libcontainers_unstable.gpg]\
+              https://download.opensuse.org/repositories/devel:kubic:libcontainers:unstable/xUbuntu_$(lsb_release -rs)/ /" \
```
dirkmueller, line 157:
```suggestion
              https://download-repositories.opensuse.org/repositories/devel:kubic:libcontainers:unstable/xUbuntu_$(lsb_release -rs)/ /" \
```

.github/workflows/ci.yaml
```diff
@@ -144,11 +144,22 @@ jobs:
         uses: actions/setup-python@v4
         with:
           python-version: '3.x'
+
       - name: Install necessary dependencies
         if: ${{ matrix.container_runtime == 'PODMAN' }}
         run: |
-          sudo apt update && sudo apt install podman
+          sudo mkdir -p /etc/apt/keyrings
+          curl -fsSL https://download.opensuse.org/repositories/devel:kubic:libcontainers:unstable/xUbuntu_$(lsb_release -rs)/Release.key \
```
dirkmueller, line 152:
```suggestion
          curl -fsSL https://download-repositories.opensuse.org/repositories/devel:kubic:libcontainers:unstable/xUbuntu_$(lsb_release -rs)/Release.key \
```

.github/workflows/ci.yaml
dcermak: I have seen so many tests fail with connection failures to download.suse.com, this _might_ help.

dcermak:
This is still not going to work as the opensuse/nginx image is signed with the key `D754694F9AB48CE9` which is whatever :shrug: but most certainly **not** the key from `openSUSE:Factory`

dirkmueller:
> This is still not going to work as the opensuse/nginx image is signed with the key `D754694F9AB48CE9` which is whatever shrug but most certainly **not** the key from `openSUSE:Factory`

This is the container signing key, aka the right one. I've updated the PR to fix this.

.github/workflows/ci.yaml
dirkmueller: I have reverted that. we should do that conciously in a separete PR if it really is still a problem

dirkmueller:
PHP tests have passed

PR #227, opened 2023-04-05
Cleanups

dirkmueller:


bci_tester/data.py
```diff
@@ -128,7 +128,7 @@ def _get_repository_name(image_type: _IMAGE_TYPE_T) -> str:
         return "containerfile/"
     if image_type == "kiwi":
         return "images/"
-    assert False, f"invalid image_type: {image_type}"
```
dcermak, line 131: Any particular reason for this change (I'm really just curious)?

dirkmueller: > Any particular reason for this change (I'm really just curious)?

Flake8 dislikes them because "assert" is removed when building with optimizations (which we do not do). The recommendation is to use AssertionError which is not removed. 

see https://pydigger.com/pypi/flake8-bugbear (search for B011)

dcermak:
It's green, merge it before it moves!

PR #234, opened 2023-04-18
Don't rebuild the postgres image to avoid HEALTHCHECK disappearing

dcermak:
For some reason buildah on RHEL 7 does not propagate HEALTHCHECK from the base image, which causes test failures as pytest_container cannot wait for a HEALTCHECK.


dcermak:
Note that this is still broken with Postgres 14

PR #235, opened 2023-04-18
Postgres on rhel7

dcermak:


jlausuch (approved):
Thanks for the fix, VR passing: https://openqa.suse.de/tests/10940183#step/bci_test_podman/40

I actually don't like we have to code these kind of hacks because of ancient version of container tools. Let's see if we can drop support for rhel7 and focus on more modern things...

dirkmueller (commented):
If we can avoid testing on rhel7 then I would prefer to avoid these kinds of hacks on the codebase and not merge it.

If we do need to continue to support rhel7 then this would require customer documentation or automatic detection (which I think might not be possible)

dcermak:
> I actually don't like we have to code these kind of hacks because of ancient version of container tools. Let's see if we can drop support for rhel7 and focus on more modern things...

This is not really that much of a hack to be honest… It's just a different way how to accomplish the same thing.

> If we do need to continue to support rhel7 then this would require customer documentation or automatic detection (which I think might not be possible)

Just one more year at most :crossed_fingers:. Jokes aside, there's imho not much to document: the buildah issue would be an upstream problem and not really something for us to fix. And that you have to fallback to a different form of auth method when your postgres version is ancient is also kinda a downstream problem as well…

jlausuch:
> So what should we do about this? @jlausuch do you want to keep the current broken state on RHEL 7 until you have the conversation with PM?

Let's merge it. RHEL 7.9 will be in general support for some time and it's our duty to support it...

PR #237, opened 2023-04-21
Remove periodic CI tasks

dcermak:
I don't have the time to review & fix them periodically. And the security scan has been broken/red since ever.


.github/workflows/ci.yaml
```diff
@@ -2,8 +2,6 @@
 name: CI
 
 on:
-  schedule:
-    - cron: '44 4 */2 * *'
```
dirkmueller, line 6: I would like to keep those. It is useful to see the state without a PR under test to compare the results with.

dirkmueller (changes requested):
I'm okay with the security scan going away. I regularly check the periodic CI so I would like to keep them.

As most of the changes are happening outside PRs it is highly useful to see new failures. I agree the random fails are annoying however thats a useful low priority task to improve those

.github/workflows/ci.yaml
dcermak: Please don't take this wrongly, but does that mean that you will act upon CI failures then? I see no value in a periodically running CI just to have it. And currently it is just running, sending out failure emails and **no one** has the time to fix them.

dcermak: I have added them back

dirkmueller (approved):

PR #238, opened 2023-04-21
Add a test that `zypper up` works in every container image

dcermak:


tests/test_all.py
```diff
@@ -101,6 +103,14 @@ def test_glibc_present(auto_container):
         assert auto_container.connection.exists(binary)
 
 
+@pytest.mark.parametrize(
+    "container_per_test", CONTAINERS_WITH_ZYPPER, indirect=True
+)
+def test_zypper_up_works(container_per_test: ContainerData) -> None:
+    """Check that :command:`zypper -n up` works."""
+    container_per_test.connection.run_expect([0], "zypper -n up")
```
jlausuch, line 111: btw, does this command have a default timeout?

jlausuch (dismissed):

dirkmueller:
Why are we doing this? What about container without zypper? I don't think this makes sense for application containers

tests/test_all.py
dcermak: I don't think it has one

dcermak:
> Why are we doing this?

This idea came up during our containers doc sync. A `zypper -n ref && zypper -n up` is a very common pattern in dockerfiles, so we should be testing that it actually works and isn't broken in some subtle way.

> What about container without zypper?

This test only runs on all containers in the list `CONTAINERS_WITH_ZYPPER`.

>  I don't think this makes sense for application containers

I don't have an opinion on this tbh.

tests/test_all.py
dirkmueller: I added a `timeout`

dirkmueller (dismissed):

PR #244, opened 2023-05-05
allow downgrade

dirkmueller:
SLE_BCI repo is usually behind when containers are tested, so the "no-downgrade" trips. allow downgrades for now until situation is improved that the BCI repo is refreshed at the same time or before the containers are.

### DIFF
--- a/tests/test_all.py
+++ b/tests/test_all.py
@@ -111,11 +111,15 @@ def test_zypper_dup_works(container_per_test: ContainerData) -> None:
     BCI repo by running :command:`zypper -n dup` and checking that there are no
     downgrades or arch changes.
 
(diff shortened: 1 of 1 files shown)


dcermak:
Are we then really testing anything useful at all?

dirkmueller:
> Are we then really testing anything useful at all?

yes, we're testing that the containers do not contain anything that would be "conflicting" with what is in SLE_BCI. It might bring a lot but maybe someting

grisu48:
Let's please add a comment stating that this is needed only due to unresolved issues with the BCI repo sync issue.

pdostal (approved):

grisu48:
Post-Merge approval. Thanks!

PR #247, opened 2023-05-11
Remove chalk from common repos

grisu48:
The chalk test is broken. Remove it from the list of packages to be tested.


dirkmueller:
I think we should xfail it rather tahn removing it.

dirkmueller:
I made this change in #249

grisu48:
Agreed, thanks for the PR!

PR #257, opened 2023-06-08
Increase timeout for zypper

grisu48:
Increase the timeout for zypper to prevent false positives on some slow architectures.

Example failues: [1](https://openqa.suse.de/tests/11265480#step/bci_test_podman/31) [2](https://openqa.suse.de/tests/11265479#step/bci_test_podman/37) [3](https://openqa.suse.de/tests/11265481#step/bci_test_podman/37)


dcermak (approved):
LGTM

PR #260, opened 2023-06-18
Switch testing to 15.5

dirkmueller:


tests/test_build.py
```diff
@@ -36,6 +40,7 @@ def test_container_build_and_repo(container_per_test, host):
     )
 
     repos = get_repos_from_connection(container_per_test.connection)
+    repo_names = [repo.name for repo in repos]
```
dirkmueller, line 43: make this a set() here?

tests/test_build.py
```diff
@@ -48,7 +53,14 @@ def test_container_build_and_repo(container_per_test, host):
             len(repos) > 1
         ), "On a registered host, we must have more than one repository on the host"
     else:
-        assert len(repos) in (1, 2)
+        # SLE_BCI (optionally SLE_BCI_debug, SLE_BCI_source & MS .Net repo)
+        assert len(repos) <= 4
+        assert not set(repo_names) - {
```
dirkmueller, line 58: so that we don't need to convert it into a set there?

tests/test_build.py
```diff
@@ -59,6 +71,21 @@ def test_container_build_and_repo(container_per_test, host):
     assert sle_bci_repo.name == "SLE_BCI"
     assert sle_bci_repo.url == BCI_DEVEL_REPO
 
+    for repo_name in "SLE_BCI_debug", "SLE_BCI_source":
+        candidates = [repo for repo in repos if repo.name == repo_name]
+        assert len(candidates) in (0, 1)
+
+        if candidates:
+            container_per_test.connection.run_expect(
+                [0], f"zypper modifyrepo --enable {candidates[0].alias}"
```
dirkmueller, line 80: this mutates the state (enables the repository). why are we doing that? what does it test? we already tested earlier that it is listed. I think we should test that it is disabled by default perhaps?

dirkmueller: enabling the repository twice doesn't throw an error so this test doesn't do that..

dcermak: The last line in this test function runs `zypper ref`, so that would fail if the url were wrong.

dirkmueller: ah, so we're testing whether the repo is accessible. that makes sense. a comment would have been helpful.

dcermak:
> @dcermak you removed the "available_versions=15.5" from all the language containers, but didn't remove 15.4 from the list of available language containers. as we are going to disable 15.4 from building and updating, why do we want to keep being able to test those?

I would keep the ability to test them until 15.5 is properly released and switch then.

tests/test_build.py
dcermak: Done

PR #262, opened 2023-06-20
Remove outdated envs

dirkmueller:


dcermak (changes requested):
This will not work, the markers are still automatically added via the build tags.

PR #265, opened 2023-06-27
Test Language and App containers only on 15.5 by default

dirkmueller:
Introduce a new DEFAULT setting that is globally applied for testing if no more specific setting is being provided. Skip testing of application and language containers on 15.4 now that we have rolled over to 15.5

### DIFF
--- a/bci_tester/data.py
+++ b/bci_tester/data.py
@@ -31,18 +31,32 @@ from bci_tester.runtime_choice import DOCKER_SELECTED
 #: places
 OS_VERSION = os.getenv("OS_VERSION", "15.5")
 
-ALLOWED_OS_VERSIONS = ("15.3", "15.4", "15.5")
-_LANGUAGE_APPLICATION_STACK_OS_VERSIONS = ("15.4", "15.5")
+# Allowed os versions for base (non lang/non-app) containers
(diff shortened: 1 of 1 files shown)


bci_tester/data.py
```diff
@@ -31,19 +31,33 @@
...
+# Test Language and Application containers by default for these versions
+_DEFAULT_NONBASE_OS_VERSIONS = ("15.5",)
+
+assert sorted(ALLOWED_BASE_OS_VERSIONS) == list(
+    ALLOWED_BASE_OS_VERSIONS
+), f"list ALLOWED_OS_VERSIONS must be sorted, but got {ALLOWED_BASE_OS_VERSIONS}"
+
+assert sorted(ALLOWED_NONBASE_OS_VERSIONS) == list(
+    ALLOWED_NONBASE_OS_VERSIONS
+), f"list ALLOWED_NONOS_VERSIONS must be sorted, but got {ALLOWED_NONBASE_OS_VERSIONS}"
```
dcermak, line 49: Why did you add this part? It's not wrong, but it's not user modifiable, so why?

dirkmueller: No particular reason, just to give hints to my future self on when the next times those settings/variables/constants are modified which rules they need to comply with.

PR #269, opened 2023-07-11
Enable regular CI for OS_VERSION=tumbleweed as well

dirkmueller:


tests/test_389ds.py
```diff
@@ -9,7 +9,7 @@
 def test_ldapwhoami(auto_container_per_test):
     auto_container_per_test.connection.run_expect(
         [0],
-        "dsconf localhost backend create --suffix dc=example,dc=com --be-name userRoot",
+        "sleep 10 && dsconf localhost backend create --suffix dc=example,dc=com --be-name userRoot",
```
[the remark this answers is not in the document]
dirkmueller: because pytest needs to wait until the container has finished starting up. we used to use HEALTHCHECK for this, but this no longer exists in OCI formats.

dcermak:
The python cryptography test case will need to be run inside a virtual env to work on tumbleweed.

dirkmueller:
> The python cryptography test case will need to be run inside a virtual env to work on tumbleweed.

my initial plan was to change the container rather than the tests

tests/test_base.py
```diff
@@ -46,9 +46,19 @@ def test_base_size(auto_container, container_runtime):
...
-        < BASE_CONTAINER_MAX_SIZE[LOCALHOST.system_info.arch] * 1024 * 1024
+        // 1024
+        // 1024
+    )
+    min_container_size = container_size - 3
+    assert (
+        min_container_size
+        < container_size
+        <= BASE_CONTAINER_MAX_SIZE[LOCALHOST.system_info.arch]
+    ), (
```
dcermak, line 59: I don't understand the logic behind this, won't `min_container_size < container_size` be always true?

tests/test_busybox.py
```diff
@@ -43,9 +43,17 @@ def test_busybox_image_size(
...
-        < size[LOCALHOST.system_info.arch] * 1024 * 1024
+        // 1024
+        // 1024
+    )
+    min_container_size = container_size - 2
+    assert (
+        min_container_size < container_size <= size[LOCALHOST.system_info.arch]
+    ), (
+        f"Base container size is {container_size} MiB for {LOCALHOST.system_info.arch} "
+        f"(expected {min_container_size}..{size[LOCALHOST.system_info.arch]} MiB)"
```
dcermak, line 56: Same question as in test_base.py. And why is there a `-2` here and not a `-3`?

tests/test_python.py
```diff
@@ -100,11 +101,19 @@ def test_pip(auto_container):
     assert f"pip {version_from_env}" in reported_version
 
 
+@pytest.mark.skipif(
+    OS_VERSION == "tumbleweed",
+    reason="pip --user not working properly with PEP 668",
```
dcermak, line 106: This sounds a bit better maybe?
```suggestion
    reason="pip --user not working properly due to PEP 668",
```

dcermak (changes requested):
Postgres & Python tests are still red

dirkmueller:
> Postgres & Python tests are still red

yes, but there are fixes for that in the works (python is a bug in packaging and postgres is the BCI-dockerfile-generator fix)

tests/test_base.py
dirkmueller: no. For tumbleweed the containers are currently about 20MB larger than they actually should be because dbus-1 etc is pulled into the container, which it doesn't belong there. I want to detect when we finally managed to get rid of the excessive dependencies again by having a failing test, and then lower the expected container size to the lower values again. 

without tihs change, we would not notice after the container bloat has been fixed that another bloat has been added. I think we do want to notice if the size of the containers change significantly (where significantly here is defined to be +/- 3MB)

tests/test_ruby.py
```diff
@@ -90,6 +90,11 @@ def test_rails_hello_world(auto_container_per_test):
 
 @pytest.mark.skipif(OS_VERSION != "tumbleweed", reason="no rails for ruby 2.5")
 def test_rails_template(auto_container_per_test):
+    # Rails asset pipeline needs Node.js and yarn
+    auto_container_per_test.connection.run_expect(
+        [0], "zypper -n in nodejs-default yarn libyaml-devel"
```
dcermak, line 95: you don't need nodejs & yarn
```suggestion
        [0], "zypper -n in ruby-devel libyaml-devel"
```

dirkmueller: so on the suggestion: ruby-devel should be already part of the container, if it isn't then that should be fixed first. a devel container for ruby without ruby-devel doesn't make any sense. 

and rails is printing warnings if it cannot compile the asset pipeline

tests/test_base.py
```diff
@@ -46,9 +46,18 @@ def test_base_size(auto_container, container_runtime):
             "ppc64le": 160,
             "s390x": 125,
         }
-    assert (
+    container_size = (
         container_runtime.get_image_size(auto_container.image_url_or_id)
-        < BASE_CONTAINER_MAX_SIZE[LOCALHOST.system_info.arch] * 1024 * 1024
+        // 1024
+        // 1024
```
dcermak, line 52: You are loosing precision here, why?

```suggestion
        / 1024
        / 1024
```

tests/test_ruby.py
dcermak: No idea how I came to the conclusion that installing `ruby-devel` would help in any way, because that package is more or less empty and only pulls in `ruby$version-devel` (but that one is already installed) :facepalm:

tests/test_base.py
dirkmueller: I've removed the min-boundary again as there are vastly different sizes depending on the version and I don't feel like tracking them per SP level for now.

PR #270, opened 2023-07-13
Fix TARGET for BCI tumblweed images

mloviska:


bci_tester/data.py
```diff
@@ -161,6 +162,8 @@ def _get_repository_name(image_type: _IMAGE_TYPE_T) -> str:
         return ""
     if TARGET == "ibs-cr":
         return "images/"
+    if TARGET == "factory-totest":
+        return "containers/opensuse/bci/"
```
dcermak, line 166: Are you sure about this? This only makes sense if all images are released into the `containers` namespace in ToTest.

dirkmueller, line 166:
```suggestion
        return "containers/opensuse/"
```

dirkmueller (changes requested):
no, this doesn't work... the /bci/ suffix is duplicated

dirkmueller (approved):

dirkmueller:
tested with `OS_VERSION=tumbleweed TARGET=factory-totest` :+1:

bci_tester/data.py
mloviska: AFAICT this applies for `init, minimal, micro, busybox, golang, rust, ruby and postgres14` images that I have checked.

mloviska:
Thank you @dirkmueller, @dcermak !

PR #272, opened 2023-07-14
Reenable FIPS tests on SLE15, skip on tumbleweed

dirkmueller:


dcermak (changes requested):
Fails with:
```
blake2b512 is not a known digest
```

tests/test_fips.py
```diff
@@ -105,20 +105,24 @@ def test_openssl_binary(
     img_id = container_runtime.get_image_id_from_stdout(cmd.stdout)
 
     exec_cmd = " ".join(
-        [container_runtime.runner_binary, "run", "--rm"]
+        [container_runtime.runner_binary, "run", "--rm", "--privileged=true"]
```
dcermak, line 108: Why do you have to run it privileged?

tests/test_fips.py
```diff
@@ -105,20 +105,24 @@ def test_openssl_binary(
...
+        host.run_expect(
+            [0],
+            f"{exec_cmd} bash -c 'mount --bind /tmp/f /proc/sys/crypto; fips-test {digest}'",
+        )
 
     for digest in NONFIPS_DIGESTS:
         err_msg = host.run_expect(
-            [1], f"{exec_cmd} /bin/fips-test {digest}"
+            [1],
+            f"{exec_cmd} bash -c 'mount --bind /tmp/f /proc/sys/crypto; fips-test {digest}'",
```
dcermak, line 122: In the CI test run, this returns 0.

dcermak (changes requested):

tests/test_fips.py
dirkmueller: because I'm running mount inside the container

tests/test_fips.py
dcermak: Now it's a 32 :face_with_spiral_eyes: 
Also: `mount: /proc/sys/crypto: mount point does not exist`

dirkmueller: I know, apparently the ubunto workers need some kernel module that doesn't autoload.

tests/test_fips.py
```diff
@@ -20,39 +20,43 @@
...
-COPY --from=builder /usr/lib64/libssl.so.1.1 /usr/lib64/
-COPY --from=builder {'/usr' if OS_VERSION not in ('15.3', '15.4') else ''}/lib64/libz.so.1 /usr/lib64/
-COPY --from=builder /usr/lib64/engines-1.1 /usr/lib64/engines-1.1
-COPY --from=builder /usr/lib64/.libcrypto.so.1.1.hmac /usr/lib64/
-COPY --from=builder /usr/lib64/.libssl.so.1.1.hmac /usr/lib64/
+COPY --from=builder /src/fips-test /usr/local/bin/fips-test
+COPY --from=builder /usr/bin/openssl /usr/bin/openssl
+COPY --from=builder /usr/lib64/libcrypto.so.* /usr/lib64/
+COPY --from=builder /usr/lib64/libssl.so.* /usr/lib64/
+COPY --from=builder /usr/lib64/libz.so.1 /usr/lib64/
```
dcermak, line 54: This will break on 15.3 & 15.4:
```
❯ podman run --rm -it registry.suse.com/bci/bci-base:15.4 ls -al /usr/lib64/libz.so.1
ls: cannot access '/usr/lib64/libz.so.1': No such file or directory
```

tests/test_fips.py
```diff
@@ -20,39 +20,43 @@
...
 
-# building the documentation will fail on a non-FIPS host otherwise
-if "sphinx" not in sys.modules:
-    assert (
-        host_fips_enabled()
-    ), "The host must run in FIPS mode for the FIPS test suite"
+#: Error message from OpenSSL when a non-FIPS digest is selected in FIPS mode
+FIPS_ERR_MSG = (
+    "not a known digest" if OS_VERSION == "15.3" else "Error setting digest"
+)
```
dcermak, line 29: The error message is now the same on all service packs

tests/test_fips.py
```diff
@@ -101,14 +105,15 @@ def test_openssl_binary(
...
-        err_msg = host.run_expect(
-            [1], f"{exec_cmd} /bin/fips-test {digest}"
-        ).stderr
+        err_msg = host.run_expect([1], f"{exec_cmd} fips-test {digest}").stderr
 
-        assert f"Unknown message digest {digest}" in err_msg
+        if Version.parse(OS_VERSION) <= Version(15, 5):
+            assert f"Unknown message digest {digest}" in err_msg
+        else:
+            assert "disabled for FIPS" in err_msg
```
dcermak, line 116: The error message is now the same on all service packs

dirkmueller:
obsolete I think

PR #274, opened 2023-07-18
Add option to target released containers

mloviska:
Run `repository` test against already released base container.

- ticket: https://progress.opensuse.org/issues/132935

### DIFF
--- a/bci_tester/data.py
+++ b/bci_tester/data.py
@@ -98,7 +98,14 @@ TARGET = os.getenv("TARGET", "obs")
 #: via this variable instead
 BASEURL = os.getenv("BASEURL")
 
-if TARGET not in ("obs", "ibs", "ibs-cr", "dso", "factory-totest"):
+if TARGET not in (
+    "obs",
+    "ibs",
+    "ibs-cr",
+    "dso",
+    "factory-totest",
+    "ibs-released",
+):
     if BASEURL is None:
         raise ValueError(
             f"Unknown target {TARGET} specified and BASEURL is not set, cannot continue"
@@ -116,6 +123,7 @@ else:
         "ibs": f"registry.suse.de/suse/{DISTNAME}/update/bci",
         "dso": "registry1.dso.mil/ironbank/suse",
         "ibs-cr": f"registry.suse.de/suse/{DISTNAME}/update/cr/totest",
+        "ibs-released": f"registry.suse.com",
     }[TARGET]
 
 
@@ -158,7 +166,7 @@ _IMAGE_TYPE_T = Literal["dockerfile", "kiwi"]
 
 
 def _get_repository_name(image_type: _IMAGE_TYPE_T) -> str:
-    if TARGET == "dso":
+    if TARGET in ("dso", "ibs-released"):
         return ""
     if TARGET == "ibs-cr":
         return "images/"


mloviska:
> 
> I don't see how running the test suite using `registry.suse.com` is related to the ticket?

Yup, it is in the suggestion section. 
`Since we want to use a stable (released) container image.`

jlausuch:
> > Run `repository` test against already released base container.
> > ```
> > * ticket: https://progress.opensuse.org/issues/132935
> > ```
> 
> I don't see how running the test suite using `registry.suse.com` is related to the ticket?

To test BCI repo. 
We have this scenarios:
- Case 1:  `SUT = BCI-repo`
- Case 2:  `SUT = BCI images`

-> Both SUTs are not released together, actually we can release a bunch of images per day, but we will never released `BCI-repo` in the same cadence, and the `BCI-repo` is not tight to a specific image, it's common for ALL images. So, they are 2 different SUTs, 2 different moving pieces.

-> Therefore, to thest the `BCI-repo`, we need to do it independently from test & release pipeline of the images. We need an independent pipeline for BCI-repo where we can automate the release in the frequency we/mgmt want.

->  To avoid testing `SUT=BCI-repo` in `SUT = BCI images` where there are 2 SUT being tested together, we consider that testing `SUT=BCI-repo` makes more sense on a "stable" platform. Stable = already tested and released.

mloviska:
##### openQA VRs

* [sle-15-SP3-BCI-Repo-Updates.x86_64](http://kepler.suse.cz/tests/21295#step/_root_BCI-tests_all_podman/1)
* [sle-15-SP4-BCI-Repo-Updates.x86_64](http://kepler.suse.cz/tests/21294#step/_root_BCI-tests_all_podman/1)
* [sle-15-SP5-BCI-Repo-Updates.x86_64](http://kepler.suse.cz/tests/21293#step/_root_BCI-tests_all_podman/1)

tests/test_repository.py
```diff
@@ -186,3 +186,26 @@ def test_sle15_packages(container_per_test, pkg):
...
+    raw = container_per_test.connection.check_output(
+        "zypper -q -s 11 pa --orphaned"
+    )
+    tokens = raw.split()
+    detected_orphans = [
+        tokens[i + 1] for i, v in enumerate(tokens) if v == "@System"
+    ]
+
+    for pkg in detected_orphans:
+        assert pkg in EXPECTED_ORPHANS
```
dcermak, line 211: Longer, but imho clearer.

```suggestion
    lines = container_per_test.connection.check_output(
        "zypper -q -s 11 pa --orphaned"
    ).splitlines()

    # output of zypper -q -s 11 pa --orphaned:
    # S   Repository  Name                  Version                Arch
    #
    # i+  @System     kubic-locale-archive  2.31-10.36             noarch
    orphans = []
    for line in lines:
        columns = line.split()
        if len(columns) == 5:
            _, location, orphan_name, _, _ = columns
            if location == "@System":
                orphans.append(orphan_name)

    assert set(orphans) == set(EXPECTED_ORPHANS)
```

mloviska: This is nice, thanks. Should we expect there will be always the same set of expected orphans. I thought that if any of those would disappear in future (for instance `kubic-locale-archive`) it should not be a problem.

dcermak: I don't know whether the orphans will stay the same. I'd just go ahead and hardcode them for now. If it turns into a problem, we can always adjust the tests later on.

dcermak (changes requested):
I don't think that the orphaned packages test does what you think it does:
```ShellSession
# zypper pa --help
packages (pa) [OPTIONS] [REPOSITORY] ...

List all packages available in specified repositories.

  Command options:

    --orphaned              Show packages which are orphaned (without repository). Default: false
```

This only checks the **locally** installed packages whether there is an update candidate.

I.e. this is a duplication of the repoclosure check and it only checks a very limited subset of packages in **one** container.

We should either add this to **all** containers or leave it out completely, as I doubt that it actually improves our test coverage.

Happy to be convinced otherwise though!

mloviska:
> I don't think that the orphaned packages test does what you think it does:
> 
> ```
> # zypper pa --help
> packages (pa) [OPTIONS] [REPOSITORY] ...
> 
> List all packages available in specified repositories.
> 
>   Command options:
> 
>     --orphaned              Show packages which are orphaned (without repository). Default: false
> ```
> 
> This only checks the **locally** installed packages whether there is an update candidate.
> 
> I.e. this is a duplication of the repoclosure check and it only checks a very limited subset of packages in **one** container.
> 

Yes, exactly. I thought that the repoclosure runs only for `dnf`.

> We should either add this to **all** containers or leave it out completely, as I doubt that it actually improves our test coverage.
> 

I thought about to move it there too, but not every container has `zypper` and we had this check only in the former repo openqa test.

> Happy to be convinced otherwise though!

dcermak:
> Yes, exactly. I thought that the repoclosure runs only for `dnf`.

repoclosure is a dnf command and it runs in a fedora container that has been configured to use only the `SLE_BCI` repository. So yes, it only runs for `dnf` by design, because zypper lacks such a command.

> I thought about to move it there too, but not every container has zypper and we had this check only in the former repo openqa test.

We have a list `CONTAINERS_WITH_ZYPPER` :-)
But I would add it only if this test was actually testing anything not already covered.

mloviska:
> > I thought about to move it there too, but not every container has zypper and we had this check only in the former repo openqa test.
> 
> We have a list `CONTAINERS_WITH_ZYPPER` :-) But I would add it only if this test was actually testing anything not already covered.

Sure, so as long as we have this case covered for `sle` I am fine to drop it :)

dcermak (approved):

jlausuch:
LGTM
Nice! This way we can completely skip the openQA code which was basically doing the same thing and rely on this test suite.

jlausuch (approved):

PR #277, opened 2023-07-27
Adjust to relocation of the serde test suite

dirkmueller:


dirkmueller:
rust tests are passing now, self approving as nobody else is there..

PR #282, opened 2023-08-08
Tumbleweed testsuite fixes

dirkmueller:


tests/test_minimal.py
```diff
@@ -7,21 +7,22 @@
...
 MICRO_IMAGE_MAX_SIZE: Dict[str, int] = {
-    "x86_64": 26,
-    "aarch64": 28,
-    "s390x": 26,
-    "ppc64le": 33,
+    "x86_64": 32 if OS_VERSION == "tumbleweed" else 26,
+    "aarch64": 34 if OS_VERSION == "tumbleweed" else 28,
+    "s390x": 32 if OS_VERSION == "tumbleweed" else 26,
+    "ppc64le": 39 if OS_VERSION == "tumbleweed" else 33,
 }
```
dcermak, line 26: Please make this two dictionaries instead

tests/test_all.py
```diff
@@ -105,9 +105,10 @@ def test_product(auto_container):
     OS_VERSION != "tumbleweed",
     reason="product flavors only available for openSUSE",
 )
-def test_opensuse_product_flavor(auto_container):
+@pytest.mark.parametrize("container", CONTAINERS_WITH_ZYPPER, indirect=True)
```
dcermak, line 108: Just out of curiosity, why is this change necessary?

dirkmueller: I tried to explain it in there commit message.

> They wouldn't be able to report the flavor anyway

It is necessary because the micro (?) container doesn't have rpm(1) die the test fails for unrelated reasons. There's no variable that lists containers with rpm, and we don't need it as the flavor only matters for containers that can and are running zypper.

PR #285, opened 2023-08-11
Add a zypper verify test

dirkmueller:
This helps to validate that we don't have unresolved dependencies in containers.


dcermak:
Great idea, thanks!

PR #290, opened 2023-08-17
Remove --retry-connrefused

grisu48:
Remove the `--retry-connrefused` option from curl, because it's not available on RES7.

* Related test failure: https://openqa.suse.de/tests/11860102#step/_root_BCI-tests_nginx_podman/1


clanig:
Thanks a lot for taking care of this!
If the option `--retry-connrefused` is removed, I assume it will exit with `7` if the service is not available and not retry any more.

Since I guess, this is the reason for the retry attempts, it would be similar to run without any retries.

I would suggest to wrapping the command itself in an own custom retry logic, not using any of the retry functionality provided by curl itself.

clanig:
LGTM.

tests/test_nginx.py
```diff
@@ -10,6 +11,11 @@ def test_nginx_welcome_page(auto_container, host):
     """test that the default welcome page is served by the container."""
     host_port = auto_container.forwarded_ports[0].host_port
 
-    assert "Welcome to nginx" in host.check_output(
-        f"curl -sf --retry 5 --retry-connrefused http://localhost:{host_port}/"
-    )
+    # Retry 5 times with quadratic backoff delay
+    retries = 5
+    for i in range(retries):
+        resp = requests.get(f"http://localhost:{host_port}/", timeout=30)
```
dirkmueller, line 17: this doesn't work, as get is throwing a requests.exceptions.ConnectionError

tests/test_nginx.py
```diff
@@ -10,6 +11,11 @@ def test_nginx_welcome_page(auto_container, host):
     """test that the default welcome page is served by the container."""
     host_port = auto_container.forwarded_ports[0].host_port
 
-    assert "Welcome to nginx" in host.check_output(
-        f"curl -sf --retry 5 --retry-connrefused http://localhost:{host_port}/"
-    )
+    # Retry 5 times with quadratic backoff delay
+    retries = 5
```
dirkmueller, line 15: please just use https://tenacity.readthedocs.io/en/latest/

tests/test_nginx.py
grisu48: Indeed, so we do need an exception handler. I wanted to avoid that

tests/test_nginx.py
```diff
@@ -10,6 +11,15 @@ def test_nginx_welcome_page(auto_container, host):
...
 
-    assert "Welcome to nginx" in host.check_output(
-        f"curl -sf --retry 5 --retry-connrefused http://localhost:{host_port}/"
+    # Retry 5 times with quadratic backoff delay
+    @retry(
+        wait=wait_exponential(multiplier=1, min=4, max=10),
+        stop=stop_after_attempt(5),
     )
+    def check_reply():
+        resp = requests.get(f"http://localhost:{host_port}/", timeout=30)
```
dirkmueller, line 20: this can be 

```patch
+        assert "Welcome to nginx" in host.check_output(
+            f"curl -sf --retry 5 http://localhost:{host_port}/"
```

instead, which avoids the requests dependency (don#t really care either way)

dirkmueller:
full tested patch

```patch

--- a/tests/test_nginx.py
+++ b/tests/test_nginx.py
@@ -1,5 +1,6 @@
 """This module contains the tests for the nginx container, the image with nginx pre-installed.
 """
+from tenacity import retry, stop_after_delay
 from bci_tester.data import NGINX_CONTAINER
 
 
@@ -10,6 +11,9 @@ def test_nginx_welcome_page(auto_container, host):
     """test that the default welcome page is served by the container."""
     host_port = auto_container.forwarded_ports[0].host_port
 
-    assert "Welcome to nginx" in host.check_output(
-        f"curl -sf --retry 5 --retry-connrefused http://localhost:{host_port}/"
-    )
+    @retry(stop=stop_after_delay(10))
+    def check_nginx_response():
+        assert "Welcome to nginx" in host.check_output(
+            f"curl -sf --retry 5 http://localhost:{host_port}/"
+        )
+    check_nginx_response()
diff --git a/tox.ini b/tox.ini
index 290d137..880cf2e 100644
--- a/tox.ini
+++ b/tox.ini
@@ -10,6 +10,7 @@ deps =
     pytest-xdist ; python_version >= "3.6"
     dataclasses ; python_version < "3.7"
     pytest-rerunfailures
+    tenacity
     typing_extensions
     git+https://github.com/dcermak/pytest_container
     doc: Sphinx
```

tests/test_nginx.py
grisu48: No preference either, @dcermak you pick.

dirkmueller: tests/test_ruby.py needs to be upated then with the same change.

grisu48: Let's first see what @dcermak picks.

tests/test_nginx.py
```diff
@@ -10,6 +14,15 @@ def test_nginx_welcome_page(auto_container, host):
...
+    # Retry 5 times with quadratic backoff delay
+    @retry(
+        wait=wait_exponential(multiplier=1, min=4, max=10),
+        stop=stop_after_attempt(5),
     )
+    def check_nginx_response():
+        resp = requests.get(f"http://localhost:{host_port}/", timeout=30)
+        resp.raise_for_status()
+        if "Welcome to nginx" not in resp.text:
+            raise ValueError("Invalid response")
```
dirkmueller, line 26: This still needs to be an assert I think?

grisu48: Don't think so - the exception should trigger a retry with tenacity, if we `assert` here then the test run would immediately fail instead of retry to fetch the contents.

I tried yesterday to get `http://localhost:12` in line 23 and the test run failed as expected.

dirkmueller: which is what it is supposed to do. it should only retry on connection refused, not on "successfully serving unexpected content"

grisu48: Now I understand you. Yes that makes more sense, PR amended.

tests/test_nginx.py
```diff
@@ -10,6 +14,14 @@ def test_nginx_welcome_page(auto_container, host):
     """test that the default welcome page is served by the container."""
```
dcermak, line 14: You can remove `host` fixture.

grisu48: done

tests/test_nginx.py
```diff
@@ -1,15 +1,27 @@
...
 
 
-def test_nginx_welcome_page(auto_container, host):
+def test_nginx_welcome_page(auto_container):
     """test that the default welcome page is served by the container."""
     host_port = auto_container.forwarded_ports[0].host_port
 
-    assert "Welcome to nginx" in host.check_output(
-        f"curl -sf --retry 5 --retry-connrefused http://localhost:{host_port}/"
+    # Retry 5 times with quadratic backoff delay
```
dcermak, line 17: not exponential?

grisu48: Leftover when playing around with tenacity. Yes, it's exponential now.

PR #291, opened 2023-08-18
improve testing of golang containers

dirkmueller:
test the race detector and validate that -openssl containers are function as well. fips tests will come later.


grisu48:
See also https://github.com/kubereboot/kured/pull/824

PR #299, opened 2023-09-05
Add softfailure for kured on non-x86_64

grisu48:
Adds a softfailure for kured, which is currently broken on non-x86_64 systems.

* Verification run [aarch64](https://openqa.suse.de/tests/12016912#step/bci_test_podman/57)

### DIFF
--- a/tests/test_go.py
+++ b/tests/test_go.py
@@ -44,6 +44,10 @@ def test_go_version(auto_container):
             repository_url="https://github.com/weaveworks/kured.git",
             repository_tag="1.13.2",
             build_command="make bootstrap-tools kured && go test -race ./...",
+            marks=pytest.mark.xfail(
+                condition=LOCALHOST.system_info.arch != "x86_64",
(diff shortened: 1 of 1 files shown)


tests/test_go.py
```diff
@@ -44,6 +44,10 @@ def test_go_version(auto_container):
             repository_url="https://github.com/weaveworks/kured.git",
             repository_tag="1.13.2",
             build_command="make bootstrap-tools kured && go test -race ./...",
+            marks=pytest.mark.xfail(
+                condition=LOCALHOST.system_info.arch != "x86_64",
+                reason="Currently broken on arch != x86_64",
```
dirkmueller, line 49: would be nice to add a link / reference so that people can check when to remove it

dirkmueller (dismissed):

tests/test_go.py
grisu48: I've added a reference to https://github.com/kubereboot/kured/issues/823

dirkmueller (approved):

PR #300, opened 2023-09-05
Add helm as golang tests now that kured turned x86_64 only

dirkmueller:


grisu48:
LGTM, But I want to do one VR for all architectures before merging. Will post the result here once ready.

dirkmueller:
@grisu48 thanks. I'm pretty sure it'll work for ppc64le as well as I have tested that in OBS before..

grisu48 (approved):
LGTM! VRs are also passing.

PR #301, opened 2023-09-06
Check that SLE_BCI repos and BCI containers are in sync

dirkmueller:


tests/test_all.py
```diff
@@ -139,7 +140,9 @@ def test_glibc_present(auto_container):
 def test_zypper_dup_works(container_per_test: ContainerData) -> None:
     """Check that there are no packages installed that we wouldn't find in SLE
     BCI repo by running :command:`zypper -n dup` and checking that there are no
-    downgrades or arch changes.
+    downgrades or arch changes. Then validate that SLE_BCI provides all the packages
```
grisu48, line 143: Isn't the `--allow-downgrade` argument in the `zypper` call in line 156 contradicting the "checking that there are no downgrades" statement? :thinking:

tests/test_all.py
```diff
@@ -150,10 +153,37 @@ def test_zypper_dup_works(container_per_test: ContainerData) -> None:
 
     container_per_test.connection.run_expect(
         [0],
-        f"timeout 2m zypper -n dup --from {repo_name} -l -d -D "
+        f"timeout 2m zypper -n dup --from {repo_name} -l "
```
grisu48, line 156:
```suggestion
        f"timeout 5m zypper -n dup --from {repo_name} -l "
```

Since we are here. `aarch64` is sometimes really slow.

tests/test_all.py
dirkmueller: yes, fixed.

tests/test_all.py
dirkmueller: done!

grisu48 (approved):
LGTM!

PR #303, opened 2023-09-06
Basalt testing

dirkmueller:


bci_tester/data.py
```diff
@@ -63,15 +70,28 @@
...
     APP_CONTAINER_PREFIX = "opensuse"
+    BCI_CONTAINER_PREFIX = "bci"
 
     #: The Tumbleweed pretty name (from /etc/os-release)
     OS_PRETTY_NAME = os.getenv(
         "OS_PRETTY_NAME",
         "openSUSE Tumbleweed",
     )
+elif OS_VERSION == "basalt":
+    OS_MAJOR_VERSION = 16
```
grisu48, line 81: Has this been agreed to be the case now?

tests/test_all.py
```diff
@@ -58,38 +58,47 @@ def test_os_release(auto_container):
...
-            # just check whether it starts with current year
-            assert (
-                int(
-                    auto_container.connection.run_expect(
-                        [0], f". /etc/os-release && echo ${var_name}"
-                    ).stdout.strip()[:4]
+        if var_name == "VERSION_ID":
+            if OS_VERSION == "basalt":
+                # Current basalt version number
+                expected_value = "0.1"
```
grisu48, line 68: Wouldn't this be better suites as a constant in `bci_tester/data.py`? :thinking:

tests/test_all.py
```diff
@@ -58,38 +58,47 @@ def test_os_release(auto_container):
...
+            elif OS_VERSION == "tumbleweed":
+                # on openSUSE Tumbleweed that is the an ever changing snapshot date
+                # just check whether it starts with current year
+                assert (
+                    int(
+                        auto_container.connection.run_expect(
+                            [0], f". /etc/os-release && echo ${var_name}"
+                        ).stdout.strip()[:4]
+                    )
+                    == datetime.datetime.now().year
```
grisu48, line 78: I'm ok with it, a better approach coulb be to use the `VERSION_ID` (e.g. `VERSION_ID="20230904"`) and ensure, that it's not older than 30 days. This would mean that we can also release on new years eve, if we feel fancy :slightly_smiling_face: 

Just a suggestion, I'm fine with how it is for now.

dirkmueller: ok, fixed

tests/test_all.py
dirkmueller: I have moved it to data.py

bci_tester/data.py
dirkmueller: this is what we use as %suse_version. I don#t know how official it is.

grisu48 (dismissed):
LGTM!

grisu48:
Last question: Is it possible to add CI test runs? I guess not yet, so that remains for our future selves to be done.

dirkmueller:
> Last question: Is it possible to add CI test runs? I guess not yet, so that remains for our future selves to be done.

there are ci test runs in this PR already? what do you mean? they're also successfully failing..

grisu48:
> there are ci test runs in this PR already? what do you mean? they're also successfully failing..

Yes, I see them now. All good from my side

PR #309, opened 2023-09-18
Remove lodash from node tests

dirkmueller:
lodash switched froom node to bun (https://bun.sh/) which is no longer node compatible.


grisu48 (dismissed):
We still have enough other packages that we test, so this should be fine.

PR #310, opened 2023-09-21
Rust 1.72 as stable

dirkmueller:


grisu48 (approved):

grisu48:
Why is the CI complaining?

PR #314, opened 2023-09-29
Add test against lifecycle data in the containers

dirkmueller:


dirkmueller:
this relies on https://bugzilla.suse.com/show_bug.cgi?id=1215834 being fixed first

PR #316, opened 2023-10-06
Add simple smoke test building drbd using the kernel-module-container

dcermak:


pyproject.toml
```diff
@@ -88,6 +88,11 @@ markers = [
     'rust_oldstable',
     'rust_stable',
     'tumbleweed_latest',
+    # not used, but required:
+    'bci-sle15-kernel-module-toolkit_15.3',
```
dirkmueller, line 92: 15.3 is not in the list of available versions above, which variant is correct?

tests/test_all.py
```diff
@@ -234,6 +235,12 @@ def test_zypper_dup_works(container_per_test: ContainerData) -> None:
         "skelcd-EULA-bci",
         "sles-release",
         "ALP-dummy-release",
+        # packages pulled into the kernel-module container:
```
dirkmueller, line 238: imho these need to be added into the SLE BCI repository as well.

tests/test_all.py
```diff
@@ -259,7 +266,13 @@ def test_zypper_verify_passes(container_per_test: ContainerData) -> None:
     [
         c
         for c in ALL_CONTAINERS
-        if (c not in [INIT_CONTAINER, PCP_CONTAINER] + POSTGRESQL_CONTAINERS)
+        if (
+            c
+            not in (
+                [INIT_CONTAINER, PCP_CONTAINER, KERNEL_MODULE_CONTAINER]
```
dirkmueller, line 272: why is systemd in the kernel container? that seems odd. plus the test's docstring needs to be uppdated

tests/test_kernel_module.py
```diff
@@ -0,0 +1,51 @@
...
+_DRBD_VERSION = "9.2.5"
+
+
+DRBD_CONTAINER = DerivedContainer(
+    base=container_from_pytest_param(KERNEL_MODULE_CONTAINER),
+    containerfile=rf"""WORKDIR /src/
+RUN zypper -n in coccinelle tar
+
+RUN set -euxo pipefail; \
+    curl -L -o drbd.tar.gz https://pkg.linbit.com//downloads/drbd/9/drbd-{_DRBD_VERSION}.tar.gz; \
```
dirkmueller, line 18: add `-sf`

tests/test_kernel_module.py
```diff
@@ -0,0 +1,51 @@
...
+    tar xvzf drbd.tar.gz; rm drbd.tar.gz; cd drbd-{_DRBD_VERSION}; \
```
dirkmueller, line 19: `tar xf` is enough

tests/test_kernel_module.py
```diff
@@ -0,0 +1,51 @@
...
+    make -C /usr/src/linux-obj/x86_64/default modules M="$(pwd)/drbd" SPAAS=false
```
dirkmueller, line 20: this is not gonna work on non-x86_64, but the test is not restricted to be run only on x86_64

tests/test_kernel_module.py
```diff
@@ -0,0 +1,51 @@
...
+
+_DPDK_VERSION = "23.07"
+
+DPDK_CONTAINER = DerivedContainer(
+    base=container_from_pytest_param(KERNEL_MODULE_CONTAINER),
+    containerfile=rf"""WORKDIR /src/
+RUN zypper -n in meson python3-pip libnuma-devel && pip install pyelftools
+
+RUN set -euxo pipefail; \
+    curl -o dpdk.tar.gz https://fast.dpdk.org/rel/dpdk-{_DPDK_VERSION}.tar.gz; tar xvzf dpdk.tar.gz; rm dpdk.tar.gz; cd dpdk-{_DPDK_VERSION}; \
```
dirkmueller, line 33: add `sf`

pyproject.toml
dcermak: I've resolved it by not adding marks to containers that are unavailable for this `OS_VERSION`

tests/test_all.py
dcermak: Done

tests/test_all.py
```diff
@@ -238,6 +239,12 @@ def test_zypper_dup_works(container_per_test: ContainerData) -> None:
         "skelcd-EULA-bci",
         "sles-release",
         "ALP-dummy-release",
+        # packages pulled into the kernel-module container:
+        "kernel-default-devel",
```
dirkmueller, line 243: lets update the BCI repo instead and then we can remove the allowlisting here.

dirkmueller (changes requested):
generally lgtm, thanks a lot. just don't want to add any further exceptions to the BCI repo test..

tests/test_all.py
dcermak: I have removed it as Riccardo & Fred have added these packages into the repo.

dirkmueller (dismissed):

dirkmueller (dismissed):

tests/test_all.py
```diff
@@ -291,7 +292,19 @@ def test_zypper_verify_passes(container_per_test: ContainerData) -> None:
...
+                [
+                    INIT_CONTAINER,
+                    PCP_CONTAINER,
+                    # kernel-module-container contains systemd due to pesign,
+                    # fixes are pending
+                    KERNEL_MODULE_CONTAINER,
+                ]
+                + POSTGRESQL_CONTAINERS
+            )
+        )
```
ricardobranco777, line 307: I would use a set here like this:

```suggestion
        if c not in {INIT_CONTAINERS, PCP_CONTAINER, KERNEL_MODULE_CONTAINER} | {POSTGRESQL_CONTAINERS}
            c
            not in (
                [
                    INIT_CONTAINER,
                    PCP_CONTAINER,
                    # kernel-module-container contains systemd due to pesign,
                    # fixes are pending
                    KERNEL_MODULE_CONTAINER,
                ]
                + POSTGRESQL_CONTAINERS
            )
        )
```

ricardobranco777 (approved):
Small suggestion, otherwise LGTM.

tests/test_all.py
dirkmueller: @ricardobranco777 this isn't working (and there  is a typo INIT_CONTAINERS->INIT_CONTAINER) because we're mixing sets and lists here (e.g. postgresql_containers is a list, so {[1,2,3]} gives

```
>>> {[1,2,3]}
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
TypeError: unhashable type: 'list'
```

ricardobranco777: > @ricardobranco777 this isn't working (and there is a typo INIT_CONTAINERS->INIT_CONTAINER) because we're mixing sets and lists here (e.g. postgresql_containers is a list, so {[1,2,3]} gives
> 
> ```
> >>> {[1,2,3]}
> Traceback (most recent call last):
>   File "<stdin>", line 1, in <module>
> TypeError: unhashable type: 'list'
> ```

You can use `set(POSTGRESQL_CONTAINERS)`.

mgrossu (approved):
LGTM

tests/test_all.py
dirkmueller: @ricardobranco777 same issue:

```
tests/test_all.py:298: in <listcomp>
    {
<string>:3: in __hash__
    ???
E   TypeError: unhashable type: 'list'
```

I don't have time to debug this right now. lets take this into a followup PR.

PR #320, opened 2023-10-13
Allowlist kernel-devel packages for test_repository

dcermak:
These got included recently for the kernel-module-container and are ok to ship in the repository

### DIFF
--- a/tests/test_repository.py
+++ b/tests/test_repository.py
@@ -6,6 +6,7 @@ from typing import Callable
 from typing import List
 
 import pytest
+from pytest_container.runtime import LOCALHOST
 
 from bci_tester.data import BASE_CONTAINER
 from bci_tester.data import OS_SP_VERSION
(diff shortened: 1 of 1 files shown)


tests/test_repository.py
```diff
@@ -144,6 +145,11 @@ def test_sle_bci_forbidden_packages(container_per_test):
         "nfs-kernel-server",
         "texlive-l3kernel.noarch",
         "purge-kernels-service.noarch",
+        f"kernel-azure-devel.{LOCALHOST.system_info.arch}",
+        "kernel-devel-azure.noarch",
+        "kernel-macros.noarch",
+        "kernel-source-azure.noarch",
+        f"kernel-syms-azure.{LOCALHOST.system_info.arch}",
```
grisu48, line 152: Shouldn't this also have a `.noarch` package or is this special?

dcermak: Apparently it's special:
```ShellSession
❯ podman run --rm -it registry.suse.com/bci/bci-base:15.5 zypper info kernel-syms-azure|grep Arch
Arch           : x86_64
```

dirkmueller (approved):

PR #322, opened 2023-10-17
Disallow downgrade

dirkmueller:
We have automated the release of SLE BCI repo, so downgrades should no longer be happening.


grisu48 (approved):

grisu48:
Please don't merge today, as I would like to walk the current container releases out of the pipeline without unexpected interruptions.

PR #324, opened 2023-10-20
Revert "Disallow downgrade"

grisu48:
Breaks all current BCI test runs.


dcermak:
test failures are due to helm getting a version bump, should be fixed via https://github.com/SUSE/BCI-tests/pull/325

dirkmueller:
I'm not finding a more precise check before midday

PR #326, opened 2023-10-20
Update helm version, 3.13.1 update was released yesterday

dirkmueller:


dcermak (changes requested):
you're missing the marker change, see #325

PR #327, opened 2023-10-20
Use check_output() where possible

dirkmueller:
check_output(..) is a shortcut for run_expect([0], ...).stdout


dcermak (approved):
Thanks, that looks much better now!

PR #329, opened 2023-10-20
replace test_zypper_dup_works with test_no_downgrade_on_install

dirkmueller:
test_zypper_dup is a bit too agressive and on the other side misses cases. so we just validate that installing any of the additional packages would not cause a downgrade in the container.

### DIFF
--- a/tests/test_all.py
+++ b/tests/test_all.py
@@ -3,8 +3,10 @@ This module contains tests that are run for **all** containers.
 """
 import datetime
 import fnmatch
+import json
 import xml.etree.ElementTree as ET
 
+import packaging.version
 import pytest
 from _pytest.config import Config
 from pytest_container import Container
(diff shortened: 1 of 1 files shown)


tests/test_all.py
```diff
@@ -196,43 +198,65 @@ def test_glibc_present(auto_container):
...
+        for solvable in system_solv["repositories"][0]["solvables"]
+    }
+    bci_pkgs = {}
+    for solvable in bci_solv["repositories"][0]["solvables"]:
+        bci_pkgs[solvable["solvable:name"]] = solvable["solvable:evr"]
+        if solvable["solvable:name"] in installed_pkgs:
+            continue
+        for req in solvable.get("solvable:requires", ()):
+            # Skip boolean dependencies or unversioned ones
+            if "(" in req or " = " not in req:
```
dcermak, line 237: This will also skip virtual requires like `pkgconfig(libfoo) = 1.2.3`.

dirkmueller: @dcermak yeah, I didn't want to implement provides/requires matching logic. so far it was finding the essential issues without that (as typically people do `Requires: %name = %version-%release` somewhere. we can always improve that later if it is necessary to find more issues (but already now it finds too many :/ )

tests/test_all.py
```diff
@@ -3,8 +3,10 @@
 """
 import datetime
 import fnmatch
+import json
 import xml.etree.ElementTree as ET
```
dcermak, line 7: This import can go now

dirkmueller: removed

tests/test_repository.py
```diff
@@ -9,11 +9,16 @@
...
 
 from bci_tester.data import BASE_CONTAINER
+from bci_tester.data import BCI_REPO_NAME
 from bci_tester.data import OS_SP_VERSION
 from bci_tester.data import OS_VERSION
 from bci_tester.data import REPOCLOSURE_CONTAINER
 
 
+_RM_ZYPPSERVICE = (
+    "rm -v /usr/lib/zypp/plugins/services/container-suseconnect-zypp"
```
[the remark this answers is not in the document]
dirkmueller: @grisu48 because otherwise the repository tests are using all repositories (e.g. the ones from the host system that was registered against SCC) rather than just the SLE_BCI repository for testing (which is the intention of the tests as we actually want to use it to gate whether we can erlease a new SLE BCI repository). the previous tests have been ineffective because they actually tested the customer repositories instead.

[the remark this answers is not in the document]
dirkmueller: I moved this part into #334

grisu48 (approved):

mgrossu (approved):

PR #332, opened 2023-10-23
Ensure that the branding packages are installed in the containers

dirkmueller:


tests/test_all.py
```diff
@@ -88,6 +88,20 @@ def test_os_release(auto_container):
         )
 
 
+def test_branding(auto_container):
+    """
+    check that the :file:`/etc/SUSE-brand` file exists and containers SLE branding
+    """
+    assert auto_container.connection.file("/etc/SUSE-brand").exists
```
dcermak, line 95: The location is `/usr/etc/SUSE-brand` on Tumbleweed.

dcermak (commented):
Do we really want to have these packages in the container? I really don't see the value add of a file with the contents:
```
SLE
VERSION = 15
```
That's just a worse `/etc/os-release`.

tests/test_all.py
dirkmueller: gah

tests/test_all.py
```diff
@@ -89,6 +89,29 @@ def test_os_release(auto_container):
...
+    reason="branding packages are known to not be installed",
+)
+@pytest.mark.parametrize(
+    "container",
+    CONTAINERS_WITH_ZYPPER,
+    indirect=True,
+)
+def test_branding(container):
+    """
+    check that the :file:`/etc/SUSE-brand` file exists and containers SLE branding
```
dcermak, line 103:
```suggestion
    check that the :file:`/etc/SUSE-brand` file exists and contains the SLE branding
```

tests/test_all.py
```diff
@@ -89,6 +89,29 @@ def test_os_release(auto_container):
...
+    """
+    location = "/etc/SUSE-brand"
+    branding = "SLE"
+    if OS_VERSION == "tumbleweed":
+        branding = "openSUSE"
+    if OS_VERSION in ("basalt", "tumbleweed"):
+        location = "/usr/etc/SUSE-brand"
```
dcermak, line 110: Why use these switches when the tests only run on Tumbleweed?

dirkmueller: it also runs on 15.6 (and it did until recently basalt completely exploded) also on basalt. eventually I hope to be able to fix it also for older versions, so the `skipif` will reduce over time (famous last words)

PR #334, opened 2023-10-25
Restrict repository checks to the BCI repo

dirkmueller:
In case the host is registered, it would otherwise use the registered channels

### DIFF
--- a/bci_tester/data.py
+++ b/bci_tester/data.py
@@ -149,6 +149,13 @@ else:
     }[TARGET]
 
 
+BCI_REPO_NAME = "SLE_BCI"
+if OS_VERSION == "tumbleweed":
+    BCI_REPO_NAME = "repo-oss"
+if OS_VERSION == "basalt":
+    BCI_REPO_NAME = "repo-basalt"
+
+
 def create_container_version_mark(
     available_versions: Iterable[str],
 ) -> MarkDecorator:
@@ -181,7 +188,7 @@ if BCI_DEVEL_REPO is None:
     BCI_DEVEL_REPO = f"https://updates.suse.com/SUSE/Products/SLE-BCI/{OS_MAJOR_VERSION}-SP{OS_SP_VERSION}/{LOCALHOST.system_info.arch}/product/"
     _BCI_CONTAINERFILE = ""
 else:
-    _BCI_CONTAINERFILE = f"RUN sed -i 's|baseurl.*|baseurl={BCI_DEVEL_REPO}|' /etc/zypp/repos.d/SLE_BCI.repo"
+    _BCI_CONTAINERFILE = f"RUN sed -i 's|baseurl.*|baseurl={BCI_DEVEL_REPO}|' /etc/zypp/repos.d/{BCI_REPO_NAME}.repo"
 
 
 _IMAGE_TYPE_T = Literal["dockerfile", "kiwi"]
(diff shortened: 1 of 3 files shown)


dirkmueller:
CI is broken due to bsc#1216578

bci_tester/data.py
```diff
@@ -544,17 +551,14 @@ def create_BCI(
 
 REPOCLOSURE_CONTAINER = DerivedContainer(
     base="registry.fedoraproject.org/fedora:latest",
-    containerfile=r"""RUN dnf -y install 'dnf-command(repoclosure)'
+    containerfile=f"""RUN dnf -y install 'dnf-command(repoclosure)'
```
dcermak, line 554: you can combine `r` and `f`, so no need for the `\` escaping then
```suggestion
    containerfile=rf"""RUN dnf -y install 'dnf-command(repoclosure)'
```

bci_tester/data.py
```diff
@@ -544,17 +551,14 @@ def create_BCI(
...
-    + BCI_DEVEL_REPO
-    + r""" \n\
-priority=100' > /etc/yum.repos.d/SLE_BCI.repo
-""",
+RUN printf '[{BCI_REPO_NAME}]\\n\
+enabled=1\\n\
+name="SLE BCI"\\n\
+autorefresh=0\\n\
+baseurl={BCI_DEVEL_REPO}\\n\
+priority=100\\n' > /etc/yum.repos.d/{BCI_REPO_NAME}.repo""",
```
dcermak, line 561:
```suggestion
RUN printf '[{BCI_REPO_NAME}] \n\
enabled=1 \n\
name="SLE BCI" \n\
autorefresh=0 \n\
baseurl={BCI_DEVEL_REPO} \n\
priority=100\n' > /etc/yum.repos.d/{BCI_REPO_NAME}.repo""",
```

tests/test_repository.py
```diff
@@ -196,7 +196,10 @@ def test_package_installation(container_per_test, pkg):
     that they are not accidentally not installable.
```
dcermak, line 197:
```suggestion
    We additionally have to remove the ``container-suseconnect`` zypper service before running the test to ensure that no SLES repositories are added on registered hosts thereby skewing our results.
```

tests/test_all.py
```diff
@@ -148,6 +149,9 @@ def test_lifecycle(auto_container):
                 support_end = datetime.datetime.strptime(
                     entry_date, "%Y-%m-%d"
                 )
+                if entry_name.startswith("go1.19-openssl"):
+                    # bsc#1216578
+                    continue
```
dcermak, line 154: How about something like this? Or is there more lifecycle data to be tested?
```suggestion
                if entry_name.startswith("go1.19-openssl"):
                    pytest.skip(reason="bsc#1216578")
```

dirkmueller: @dcermak that would abort the loop, we still want the other entries to be tested. I'd have to create a dict with fails and then diff that with the known exceptions. 

Not sure it is worth the effort, but I can do it. I don't plan on adding additional exceptions, I'd rather push people to provide correct data (and get QA to test stuff before releasing it)

bci_tester/data.py
dirkmueller: whats the point of that additional trailing whitespace? what am I missing?

bci_tester/data.py
dirkmueller: yeah, don't particularly like raw strings, but I'm not gonna argue about it..

tests/test_all.py
dcermak: then nevermind

bci_tester/data.py
dcermak: Just to reduce the diff noise. No point beyond that, so just ignore it.

dcermak (approved):
LGTM, but the CI needs https://github.com/SUSE/BCI-tests/pull/336

PR #344, opened 2023-11-16
Fix fips tests

dcermak:


dirkmueller:
jftr I had similar fixes in https://github.com/SUSE/BCI-tests/pull/272. will rebase that now.

PR #346, opened 2023-11-28
Enable LTSS SP3 base container testing generally

dirkmueller:


dirkmueller:
not sure what to do here. the containers are not on the devel project because they're not public, so we need to rethink the whole testing logic.

dirkmueller:
this is now superseded with #356 and related

PR #350, opened 2023-12-07
Verify that the expected groups are created in the init container

dirkmueller:


dcermak:
`systemd-network` is not in the Tumbleweed init container

dirkmueller:
well, the question if thats a testsuite problem or a container problem

dirkmueller:
I've removed it for now

dcermak (changes requested):
Fails on Tumbleweed with:
```
getent group systemd-timesync
```
That group also does not appear to exist in the distro anymore:
```
dnf rq --whatprovides "group(systemd-timesync)"
```
finds nothing.

dirkmueller:
it's not clear this serves any value. lets not bother with it.

PR #355, opened 2023-12-20
Add LTSS containers to the metadata tests

dcermak:
This was not done properly in https://github.com/SUSE/BCI-tests/pull/352


grisu48 (approved):

tests/test_metadata.py
```diff
@@ -187,6 +188,7 @@ def _get_container_label_prefix(
         if LOCALHOST.system_info.arch == "x86_64"
         else []
     )
+    + [(cont, "base", ImageType.OS_LTSS) for cont in LTSS_CONTAINERS]
```
dirkmueller, line 191: they're not just called base, there is also base-fips. we need a different solution

dirkmueller:
see #356  for a more complete fixup, but it still fails

PR #356, opened 2023-12-20
Fixup metadata tests for LTSS containers

dirkmueller:


dirkmueller:
(all validated manually on a system without the broken conmon)

dcermak (changes requested):
The base tests are broken because they import the now gone variable `LTSS_CONTAINERS`. Same issue breaks the documentation.

dcermak:
The SP4 LTSS tests are all red as the base-sp4-ltss image cannot be found. @dirkmueller Is your token valid for SP4 as well or has the image not been released yet?

dirkmueller:
> The SP4 LTSS tests are all red as the base-sp4-ltss image cannot be found. @dirkmueller Is your token valid for SP4 as well or has the image not been released yet?

the image is released. I guess we need a better token. it was on my todo for today.

grisu48 (approved):

PR #361, opened 2023-12-22
Skip BCI repo specific tests on LTSS containers

dirkmueller:
They don't have a SLE_BCI repo anymore, so the tests can never pass


grisu48 (dismissed):
Yeah, this should work.

PR #366, opened 2024-01-03
Create the marks for default_repos for Base containers again

dirkmueller:
This is a regression from 967db993afe3e1349538e9fde8610324e319406e which broke BCI repository pipeline


mloviska (approved):
LGTM, `test_all` is passing now

PR #372, opened 2024-01-05
Add test for pipx (jsc#PED-5573)

dirkmueller:


tests/test_python.py
```diff
@@ -87,6 +87,30 @@ def test_python_version(auto_container):
...
 
+@pytest.mark.parametrize(
+    "container_per_test", CONTAINER_IMAGES_T2, indirect=["container_per_test"]
+)
+def test_pipx(container_per_test):
+    """Test that we can install xkcdpass via :command:`pipx`."""
+
+    # the pipx situation on SLE 15 is a bit complicated
+    if not container_per_test.connection.exists("pipx"):
+        pytest.skip("pipx not available")
```
dcermak, line 98: I think it would make sense to skip this whole test via a function decorator. This way we pay the launch prize on SLE 15 only to check that pipx is not there.

dirkmueller: launching containers is not particularly expensive, and the pipx availability is changing over time. but I'll see what I can do about a decorator that needs separate maintenance of exclusions and inclusions over time.

dcermak (approved):

PR #373, opened 2024-01-06
Test Python 3.12 container

dirkmueller:


dcermak (commented):
LGTM in principle, but I'm worried about adding this without being able to test for sp6 yet. Or is this required for enabling SP6?

PR #375, opened 2024-01-09
Add --break-system-packages flag to pip install in Rancher Dockerfile

dcermak:
This is necessary as in Tumbleweed pip will no longer install packages system wide


dirkmueller (changes requested):
no, I want to get that fixed in rancher instead: https://github.com/rancher/rancher/pull/43868

reportedly it is just waiting for QA

PR #379, opened 2024-01-10
Disable Rancher build tests on s390x

dcermak:
rancher/machine dropped s390x support with https://github.com/rancher/machine/pull/234


mloviska (approved):
Thanks!

PR #380, opened 2024-01-10
SP6 repoclosure fixes

dcermak:


tests/test_repository.py
```diff
@@ -39,13 +39,19 @@
...
             "typelib-1_0-Gtk-4_0",
-            "python311-aiohttp",
-            "python311-libcst",
             "kernel-default-devel",
         ]
         if OS_SP_VERSION >= 4
         else []
     )
+    + (
+        # these packages are no longer in SLE 15 SP6
```
dirkmueller, line 48: they should be in SP6, this is a bug (and likely because the BCI repo pipeline is not working)

dirkmueller (changes requested):
see inline comment. I wonder if it would make sense to work on that in teh SP6 enbelement PR #359 instead

PR #384, opened 2024-01-12
Add MariaDB client container to ALL_CONTAINERS 

dcermak:
We forgot to add it and now QA's tests are failing as `tox -e all -- -k mariadb_client-10.6` collects 0 jobs :see_no_evil:


mgrossu (approved):

dirkmueller (approved):
I had done this I swear. no idea how I undid that :/

PR #393, opened 2024-02-06
Test the MS rpm repo on all .Net container images, not just the SDK

dcermak:
We are adding the repository into the containers and hence we should be testing it as well.


dirkmueller:
The test fails on containers not running as root, and eventually I think we'd like to switch over to non-root containers like upstream does. hence the change that you're reverting here. we could add  a condition if we're running as root for example. I however don't understand the value of testing a repository that you can not actually make use of.

PR #396, opened 2024-02-09
Switch golang from 1.20 & 1.21 to 1.21 & 1.22

dcermak:


bci_tester/data.py
```diff
@@ -415,8 +415,8 @@ def create_BCI(
             extra_marks=[pytest.mark.__getattr__(f"golang_{stability}")],
         )
         for golang_version, stability in (
-            ("1.20", "oldstable"),
-            ("1.21", "stable"),
+            ("1.21", "oldstable"),
```
dirkmueller, line 418: I wonder if we shouldn't finally go with oldstable and stable to not have to do that every 6 months

dirkmueller:
alternatively in #397  - the bci repo failure is handled via a repo update

PR #399, opened 2024-02-13
Ensure that all .Net tests have the marks from the original container

dcermak:


mgrossu:
VRs: https://openqa.suse.de/tests/13510383, https://openqa.suse.de/tests/13510384

mgrossu (approved):

PR #403, opened 2024-02-24
Ci fixes

dirkmueller:


dirkmueller:
tests previously failed, now they pass. I guess that's a good thing.

.github/workflows/ci.yaml
```diff
@@ -349,7 +349,7 @@ jobs:
           done
 
       - name: Run tox job
-        run: sudo --preserve-env tox -e ${{ matrix.toxenv }} -- -n auto --reruns 3
+        run: sudo --preserve-env -H tox -e ${{ matrix.toxenv }} -- -n 3 --reruns 3
```
dcermak, line 352: Why the `-n 3`? I had the impression that most of the github action workers had only two cores

PR #416, opened 2024-03-13
Run test suite as non-root

dcermak:


.github/workflows/ci.yaml
```diff
@@ -348,9 +348,17 @@ jobs:
...
-        run: sudo --preserve-env -H tox -e ${{ matrix.toxenv }} -- -n 3 --reruns 3
+      - name: Run the tests
+        run: tox -e ${{ matrix.toxenv }} -- -n auto --reruns 3 --pytest-container-log-level=debug
         env:
           CONTAINER_RUNTIME: ${{ matrix.container_runtime }}
           OS_VERSION: ${{ matrix.os_version }}
           TARGET: ${{ matrix.testing_target != '' && matrix.testing_target || 'obs' }}
+
+      - name: Run tests as root for podman
+        run: sudo --preserve-env tox -e ${{ matrix.toxenv }} -- -n auto --reruns 3 --pytest-container-log-level=debug
```
dirkmueller, line 359: why did you remove the "-H" parameter which is needed to set the home directory so that pip caching works?

dcermak: that was a mistake

.github/workflows/ci.yaml
```diff
@@ -348,9 +348,17 @@ jobs:
...
-        run: sudo --preserve-env -H tox -e ${{ matrix.toxenv }} -- -n 3 --reruns 3
+      - name: Run the tests
+        run: tox -e ${{ matrix.toxenv }} -- -n auto --reruns 3 --pytest-container-log-level=debug
         env:
           CONTAINER_RUNTIME: ${{ matrix.container_runtime }}
           OS_VERSION: ${{ matrix.os_version }}
           TARGET: ${{ matrix.testing_target != '' && matrix.testing_target || 'obs' }}
+
+      - name: Run tests as root for podman
+        run: sudo --preserve-env -H tox -e ${{ matrix.toxenv }} -- -n auto --reruns 3 --pytest-container-log-level=debug
```
dirkmueller, line 359: the `-n 3` was intentional because `-n auto` triggers too many parallel pulls which is hitting rate limiting of the CDN and then causes testcases to fail

dirkmueller (dismissed):

conftest.py
```diff
@@ -63,6 +63,17 @@ def container_git_clone(request: SubRequest, tmp_path):
...
     )
+    # fix file permissions for the git copied git repo
+    check_output(
+        [
+            runtime.runner_binary,
+            "exec",
+            container_fixture.container_id,
+            "/bin/sh",
+            "-c",
+            f"chown --recursive $(id -u):$(id -g) {git_repo_build.repo_name}",
```
dirkmueller, line 74: there gotta be a better way than this... is the issue that we'r enot setting a home directory on the container?

tests/test_go.py
```diff
@@ -67,19 +67,22 @@ def test_build_kured(auto_container_per_test, container_git_clone):
     [
         GitRepositoryBuild(
             repository_url="https://github.com/helm/helm.git",
-            repository_tag="v3.14.2",
-            build_command="make build test-unit",
+            repository_tag="v3.14.3",
+            build_command="make build && make test-unit",
```
dirkmueller, line 71: where's the difference between the two? as we don't run with "-j" it should be the same?

conftest.py
dcermak: No the issue is that `docker cp ` operates in archive mode by default preserving the uid & gid of the git repository. So the cloned repo has everything owned by uid 1000 (well, on my box, but generally != 0) and the process in the container is run under uid 0. And with that anything invoking git on that repo fails due to an ownership missmatch.

tests/test_go.py
dcermak: actually none, I thought it would fix the test issues before looking closer

dirkmueller (approved):

PR #422, opened 2024-03-25
Validate registry.suse.com containers as well

dirkmueller:
The signing key has changed and publishing was half broken. so lets test for it so that we find it immediately

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -221,6 +221,10 @@ jobs:
         run: |
           cat << EOF | sudo tee /etc/containers/registries.d/opensuse.yaml
           docker:
+            registry.suse.de:
+              use-sigstore-attachments: true
+            registry.suse.com:
+              use-sigstore-attachments: true
             registry.opensuse.org:
               sigstore: https://registry.opensuse.org/sigstore
           EOF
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yaml
```diff
@@ -326,6 +330,28 @@ jobs:
...
+          +hIxtmPOhqDKHH0As113sjTqqo20Ik233x9FFeTFD8Or7ahpqjiv5YCufk9AoQbC
+          xMIjrK9RkQYgW4RycgvXGASobwN8EE+OsMcyMUER/pdCtQhTQCc1jYLt85VhfEkC
+          4k9szMB8eZrdV9re/Ku6vnCeXRR5yn2NWKO88U4HfxEpJv5s2uFJi37+x/v9w7Uh
+          +864W/9NexXg/JFNsvh0Kmxsbi3ZegaouLyrMCHwSA3ByBZ2yCf2VuFPyUCNEZOH
+          Owi0oc9TgY1yopjsTneyGaMCAwEAAQ==
+          -----END PUBLIC KEY-----
+          EOF
+
+          cat /etc/containers/policy.json | jq .
+
```
dcermak, line 354: Debugging leftover?
```suggestion
```

dcermak:
Looks like SP3 & SP4 have not yet been re-signed with the old key. If they won't be resigned, then we'll have to resort to a different key for them.

.github/workflows/ci.yaml
dirkmueller: yeah, this is still a draft..

dirkmueller:
> Looks like SP3 & SP4 have not yet been re-signed with the old key. If they won't be resigned, then we'll have to resort to a different key for them.

the plan is to fix the signing. That's why I added tests. the tests can be merged when they're green to avoid regressing on this in the future.

dirkmueller:
remaining failures are unrelated to this change

PR #423, opened 2024-03-25
tests/test_repository.py: Move from dnf repoclosure to zypper + installcheck

Vogtinator:
Avoids workarounds and dependence on a fedora container.


tests/test_repository.py
```diff
@@ -13,79 +13,20 @@
...
-
 
 def get_package_list(con) -> List[str]:
     """This function returns all packages available from the ``SLE_BCI`` repository
     given a container connection.
 
     """
     package_list = con.check_output(
-        f"dnf list --available|grep -F '{BCI_REPO_NAME}' | cut -d' ' -f1",
+        f'zypper --xmlout se -r SLE_BCI | sed -n \'s/^<solvable.*name="\\([^"]*\\)".*$/\\1/p\'',
```
dirkmueller, line 29: I think in some other places I did the xml parsing in python to avoid messing with shell substitutions.

tests/test_repository.py
```diff
@@ -112,37 +53,19 @@ def f(pkg_name: str) -> bool:
...
-                REPOCLOSURE_FALSE_POSITIVES + KNOWN_BROKEN
-            ),
-            get_package_list(container_per_test.connection),
-        )
+@pytest.mark.parametrize("container_per_test", [BASE_CONTAINER], indirect=True)
+def test_installcheck(container_per_test):
+    """Run installcheck against the SLE_BCI repo + locally installed packages."""
+    container_per_test.connection.run_expect(
+        [0],
+        "zypper ref && installcheck $(uname -m) --exclude 'rpm-ndb' /var/cache/zypp/solv/{SLE_BCI,@System}/solv",
```
dirkmueller, line 61: why are we checking `@System/solv` ? the packages wouldn't be installed if they fail to install. if its just to fulfill dependencies, then add it after a `--nocheck` parameter. 

but imho it could be all left out alltogether. 

also, to be debated, it should be run `--withobsoletes` which will throw a bunch of errors that we imho should be cleaning up

Vogtinator: > why are we checking @System/solv ? the packages wouldn't be installed if they fail to install. if its just to fulfill dependencies, then add it after a --nocheck parameter.

Can do.

> but imho it could be all left out alltogether.

No, then everything fails because of aaa_base -> distribution-release.

> also, to be debated, it should be run --withobsoletes which will throw a bunch of errors that we imho should be cleaning up

IMO for a later step

tests/test_repository.py
```diff
@@ -205,15 +128,11 @@ def test_sle_bci_forbidden_packages(container_per_test):
     OS_VERSION not in ALLOWED_BCI_REPO_OS_VERSIONS,
     reason="no included BCI repository - can't test",
 )
-@pytest.mark.parametrize(
-    "pkg", ["git", "curl", "wget", "unzip"] + REPOCLOSURE_FALSE_POSITIVES
-)
+@pytest.mark.parametrize("pkg", ["git", "curl", "wget", "unzip"])
```
dirkmueller, line 131: nit: this could be tuple rather than a list

tests/test_repository.py
Vogtinator: Can do. Parsing xml with sed is not great, but then again zypper outputs pseudo-xml anyway...

tests/test_repository.py
dirkmueller: ok, then we should add it after --nocheck, something like

```
zypper ref && cd /var/cache/zypp/solv && installcheck $(uname -m) --exclude .. SLE_BCI.solv --nochecks System.solv```

File: tests/test_repository.py (hunk 1 above)
Comment (Vogtinator) [reply tests/test_repository.py]: Done

File: tests/test_repository.py (hunk 2 above)
Comment (Vogtinator) [reply tests/test_repository.py]: Done

**Review by dirkmueller — COMMENTED** [review]
code wise this change is fine by me. I am not 100% certain that repocheck had no other additional usecases over installcheck though. 

from the chat we had yesterday I however understood that we add installcheck into the bci repo publishing pipeline not remove it from bci-tests? what am I missing?

File: tests/test_repository.py [hunk 4]
```diff
@@ -114,37 +56,19 @@ def f(pkg_name: str) -> bool:
...
-            package_name_filter_func(
-                REPOCLOSURE_FALSE_POSITIVES + KNOWN_BROKEN
-            ),
-            get_package_list(container_per_test.connection),
-        )
+@pytest.mark.parametrize("container_per_test", [BASE_CONTAINER], indirect=True)
+def test_installcheck(container_per_test):
+    """Run installcheck against the SLE_BCI repo + locally installed packages."""
+    container_per_test.connection.run_expect(
+        [0],
```
dcermak, line 63:
```suggestion
    container_per_test.connection.check_output(
```

tests/test_repository.py
```diff
@@ -114,37 +56,19 @@ def f(pkg_name: str) -> bool:
...
+        "zypper ref && installcheck $(uname -m) --exclude 'rpm-ndb' /var/cache/zypp/solv/SLE_BCI/solv --nocheck /var/cache/zypp/solv/@System/solv",
```
dcermak, line 64: Could you add a short explanation what these two paths are and why one of them is behind `--nocheck`? I am not familiar with how installcheck works

Vogtinator: Done

tests/test_repository.py
Vogtinator: Looks like it got applied already?

dcermak (approved):

PR #428, opened 2024-03-26
Minimal Fix for the new SP6 repoclosure failures

dirkmueller:
Another boolean dependency that repoclosure just doesn't handle


tests/test_repository.py
```diff
@@ -51,6 +51,8 @@
             "jeos-firstboot",
             # has a boolean dependency on the kernel
             "ecryptfs-utils",
+            # has a boolean dependency on xwayland
+            "at-spi2-core-2",
```
dcermak, line 55: The package name is different:
```suggestion
            # has a boolean dependency on xwayland
            "at-spi2-core",
```

PR #429, opened 2024-03-26
Use the full registry url for the git server container

dcermak:
The `GIT_SERVER_CONTAINER` was implicitly relying on `bci/bci-base` resolving to `registry.suse.com/bci/bci-base`. This is implemented via `shortnames.conf`, but the respective change never made it into SLES 15 SP2. Instead we use the full URL and will use the latest bci-base for now.


ricardobranco777 (approved):
VR passed: https://openqa.suse.de/tests/13874629

PR #430, opened 2024-03-27
Test README.md urls in the container annotations (Fixes #427)

### LINKED ISSUE #427: Test container READMEs
Container have to have a pointer to a README.md. we should be checking for that.

Ideally we have a way to fetch the reference and validate its structure with a markdown linter. Spell checking for common errors could be a stretch goal but isn't required

dirkmueller:


tests/test_metadata.py
```diff
@@ -375,6 +378,50 @@ def test_disturl(
...
+    verify that if it exists, it is reachable and the content is valid
+    markdown.
+    """
+    labels = container.inspect.config.labels
+
+    readme_url = urlparse.urlparse(labels["io.artifacthub.package.readme-url"])
+
+    assert readme_url.scheme == "https"
+    assert readme_url.netloc in (
+        "api.opensuse.org",
```
dcermak, line 402: This one shouldn't be allowed, it's only reachable with credentials, which makes it useless for the use case of artifacthub

dirkmueller: api.opensuse.org has a /public route that does not require credentials and can be used for the use case of artifacthub

dcermak (changes requested):
I do not think that this is worth testing:
1. markdown consistency should be enforced in the BCI-dockerfile-generator and not here
2. the readme-url for non-tumbleweed is useless in the staging as it points to OBS/IBS and is actually incorrect (you get stuff like `https://build.opensuse.org/package/show/devel:BCI:SLE-15-SP6/nodejs-20-image?rev=95dac3af7bb409598dde88b559cfa788/README.md` which OBS cannot correctly resolve)
3. artifacthub will barf at us, if either the readme-url is wrong or the readme is not markdown
4. we didn't have a single regression/issue here

If we start seeing issues with the Readmes, then we can reconsider, but atm, this is just another test that will provide very limited value (as we **cannot** run it against staging) and be potentially yet another thing that can flake

PR #436, opened 2024-04-04
Pass output of dumpsolv -j via a temporary bind mount instead of via stdout

dcermak:
This omits these huge json files from the testinfra log, which make the output of `pytest -vv` too big for any meaningful consumption

### DIFF
--- a/tests/test_all.py
+++ b/tests/test_all.py
@@ -4,15 +4,19 @@ This module contains tests that are run for **all** containers.
 import datetime
 import fnmatch
 import json
+import pathlib
 import xml.etree.ElementTree as ET
 
 import packaging.version
 import pytest
 from _pytest.config import Config
 from pytest_container import Container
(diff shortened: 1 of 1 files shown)


dirkmueller:
is there no equivalent of `podman cp` in pytest_container? I was looking for that but couldn't find it. then we wouldn't need the bind mount but could simply copy out the file..

dcermak:
> is there no equivalent of `podman cp` in pytest_container?

No, there is not, so far there was no need for this.

> Then we wouldn't need the bind mount but could simply copy out the file..

That would be an alternative, but I am not sure if the resulting code would be really that much better. Now instead of setting up the bind mount, we'd have to copy the file into a temporary directory instead. It will be a bit shorter, but I have my doubts that it will be faster

alexandrevicenzi (approved):

PR #440, opened 2024-04-08
Remove workaround for missing java provides on java 21

dirkmueller:


tests/test_minimal.py
```diff
@@ -12,10 +12,10 @@
...
 SLE_MINIMAL_IMAGE_MAX_SIZE: Dict[str, int] = {
-    "x86_64": 49,
-    "aarch64": 51,
-    "s390x": 49,
-    "ppc64le": 59,
+    "x86_64": 49 if OS_VERSION != "15.6" else 55,
+    "aarch64": 51 if OS_VERSION != "15.6" else 56,
+    "s390x": 49 if OS_VERSION != "15.6" else 55,
+    "ppc64le": 59 if OS_VERSION != "15.6" else 65,
 }
```
dcermak, line 19: Can you please xfail this test on SP6 instead? It's a known regression that should be fixed soon, so raising the limit is imho the wrong thing as we will not catch future regressions then

tests/test_minimal.py
```diff
@@ -41,6 +41,10 @@
...
 
 
+@pytest.mark.xfail(
+    OS_VERSION == "15.6",
+    reason="known issue: rpm has too many dependencies in SLE 15.6",
```
dcermak, line 46: Looks like the faulty update got into 15.5 as well :-/

msmeissn: yes.

but I have a fixup in the pipeline

dirkmueller: is it also a problem in the base image, so do we need to fix it in 15.4 ?

dirkmueller:
passes now, lets remove test failures that block publishing of containers..

PR #441, opened 2024-04-09
Remove basalt

dirkmueller:


dcermak (approved):
LGTM, but please adjust the branch protection settings as well, now the CI is reporting that all required checks are missing

PR #442, opened 2024-04-09
[do not merge] Ci debugging

dirkmueller:


.github/workflows/ci.yaml
```diff
@@ -321,6 +322,9 @@ jobs:
           REGISTRY_LOGIN_USERNAME: ${{ secrets.REGISTRY_LOGIN_USERNAME }}
           REGISTRY_LOGIN_PASSWORD: ${{ secrets.REGISTRY_LOGIN_PASSWORD }}
 
+      - name: podman info
+        run: podman info
```
dcermak, line 326: The slow run is as root, the settings will be different
```suggestion
        run: |
          podman info
          sudo podman info
```

dirkmueller: included

PR #443, opened 2024-04-09
Use ruff for formatting the code

dirkmueller:
This gives us a bit of alignment with BCI-dockerfile-generator. Like if both would be part of the same project. sort of.

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -51,8 +51,7 @@ deps =
 [testenv:format]
 allowlist_externals = ./format.sh
 deps =
-    black < 24.1.0
-    reorder-python-imports
+    ruff
 commands =
     ./format.sh []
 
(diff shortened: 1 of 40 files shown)


dirkmueller:
this breaks the isort part of it, which in ruff is in linting, so we need to switch linting as well.

tox.ini
```diff
@@ -49,12 +49,10 @@ deps =
...
 
 [testenv:format]
-allowlist_externals = ./format.sh
 deps =
-    black < 24.1.0
-    reorder-python-imports
+    ruff
 commands =
-    ./format.sh []
+    ruff format []
```
dcermak, line 55: This will run the isort equivalent, but it'll break `--check`. Maybe resurrect `format.sh` to handle that?
```suggestion
    ruff format []
    ruff check --fix
```

dirkmueller: yeah, you could run "ruff check --diff" to figure out if there is a need to resort includes without doing that actually. I'll update the PR later. first other things.

alexandrevicenzi (approved):

PR #444, opened 2024-04-09
Remove 3.10, add 3.12

dirkmueller:
Ubuntu 22.04 is the only one with 3.10, we replace that with 24.04 which uses 3.12

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -66,7 +66,12 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        python_version: ["3.6", "3.9", "3.10", "3.11", "3.12"]
+        # 3.6: SLE15
+        # 3.9: RHEL/Centos/Liberty 9
+        # 3.11: Tumbleweed
+        # 3.12: Ubuntu 24.04
+        # 3.13: Future Tumbleweed
+        python_version: ["3.6", "3.9", "3.11", "3.12"]
     container:
       image: registry.suse.com/bci/python:${{ matrix.python_version }}
     steps:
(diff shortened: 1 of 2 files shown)


.github/workflows/ci.yaml
```diff
@@ -48,7 +48,11 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        python_version: ["3.6", "3.9", "3.10", "3.11"]
+        # 3.6: SLE15
+        # 3.9: RHEL/Centos/Liberty 9
+        # 3.11: Tumbleweed
+        # 3.12 Ubuntu 24.04
```
dcermak, line 54: The OCD is kicking in :stuck_out_tongue: 

```suggestion
        # 3.12: Ubuntu 24.04
```

dcermak (dismissed):
LGTM, but please adjust the branch protection rules again

dirkmueller:
> LGTM, but please adjust the branch protection rules again

we first need to convert the tox runs to use python 3.12 instead of 3.10 before merging this

tox.ini
```diff
@@ -17,6 +19,7 @@ deps =
     doc: Sphinx
 
 [testenv]
+basepython = python3.11
```
dcermak, line 22: Any particular reason why you are hardcoding the tests to use python 3.11? That will cause a giant headache for QA, who are using the system python in openQA.

dirkmueller: this is still a draft, I'm trying to find a workable path forward to get it to not use the system python. however, for SLE15 I don't see a problem with requiring python311 for the QA team. it might be more difficult for the tests on non-SLE hosts, but one problem at a time.

.github/workflows/ci.yaml
```diff
@@ -179,13 +184,15 @@ jobs:
...
       - name: Set up Python
         uses: actions/setup-python@v5
         with:
           python-version: "3.x"
-
+      - name: Install tox
+        run: |
+          python3 --version
+          python3 -m ensurepip
+          python3 -m pip install tox
```
dcermak, line 195: This only installs tox for the user but not for the superuser and hence line 393 with `sudo ... python3 -m tox …` fails

dcermak (changes requested):

.github/workflows/ci.yaml
dirkmueller: Yes, .. this is a draft, it doesn't require review/request changes.

dcermak:

> @dirkmueller commented on this pull request.
>
>
>
>> +          python3 --version
> +          python3 -m ensurepip
> +          python3 -m pip install tox
>
> Yes, .. this is a draft, it doesn't require review/request changes. 

Then why did you request a review yesterday?

dirkmueller:
> >Yes, .. this is a draft, it doesn't require review/request changes.
> Then why did you request a review yesterday?

I tried to clear the previous review status, which somehow was triggering a re-request. anyhow, I think it's good now.

PR #446, opened 2024-04-10
Enable pytest duration

alexandrevicenzi:
This allows to pin down what is slowing our CI.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -59,7 +59,7 @@ jobs:
           python3 --version
           python3 -m ensurepip
           python3 -m pip install tox
-      - run: 'tox -e py$(echo $PY_VER | tr -d . )-unit -- -n auto'
+      - run: 'tox -e py$(echo $PY_VER | tr -d . )-unit -- -n auto --durations=25 --durations-min=600.0'
         env:
           SETUPTOOLS_SCM_PRETEND_VERSION: 1.2.3
           PY_VER: ${{ matrix.python_version }}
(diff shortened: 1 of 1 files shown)


dcermak:
> Given that it doesn't seem to have significant overhead, I wonder if we should add `--durations-min=20` or some other cutoff value permanently to have this data continuously going forward? Would help comparing individual tests between one distro and another

We have to be careful with that as QA runs the tests exactly like we do via tox, only that they also execute them in openQA, which could be substantial slower (especially on foreign architectures with emulated workers).

dirkmueller:
I think it wouldn't hurt QA either (but the extra argument can also be moved directly into the github workflow file)

alexandrevicenzi:
I find it useful to have the complete execution time, but we can print only the slowest ones. Not sure if capturing the time adds too much overhead to tests.

dirkmueller:
@alexandrevicenzi can you please update the PR accordingly so that we can review it?

alexandrevicenzi:
@dcermak @dirkmueller can we agree on a solution?

Discussion so far:

My change prints the time for all test executions.
Dirk proposed a cutoff.
Dan says we need to be careful with QA.

Taking all into account:

Print only the time execution of slow tests.
Leave the change in the workflow file, not in the tox.ini file.

dcermak:
Alexandre Vicenzi ***@***.***> writes:

> Taking all into account:
>
> Print only the time execution of slow tests.
> Leave the change in the workflow file, not in the tox.ini file.

This sounds like a good compromise to me

alexandrevicenzi:
@dcermak @dirkmueller as our tests use `-vv` the duration cutoff is ignored, the unit tests do not use `-vv`, and it works.

`(12 durations < 600s hidden.  Use -vv to show these durations.)`

If we remove `-vv` it should work, otherwise the cutoff will always be ignored in most tests.

dirkmueller (approved):

PR #448, opened 2024-04-10
Skip rancher from building

dirkmueller:
the build is completely messed up, even going back to an older release no longer works. disable it for now

### DIFF
--- a/tests/test_go.py
+++ b/tests/test_go.py
@@ -130,6 +130,7 @@ def test_build_generics_cache(
     )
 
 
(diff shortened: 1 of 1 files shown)


alexandrevicenzi (approved):

dcermak (changes requested):
iirc one of our biggest requirements was to **not** break the Rancher build. Without the test, we're completely blind. If this requirement changed, then just `rm -rf` the test, it's a giant PITA

dirkmueller:
well it xfail succeeded now, so I guess it was resolved

PR #449, opened 2024-04-10
Go back to preinstalled podman

dirkmueller:


.github/workflows/ci.yaml
```diff
@@ -162,8 +163,9 @@ jobs:
           sudo apt-get -qq -y install podman buildah
           sudo mkdir -p /etc/containers/registries.d/
 
-      - name: configure podman to validate sigstore signatures
+      - name: configure podman use sigstore attachments
```
dcermak, line 166:
```suggestion
      - name: configure podman to use sigstore attachments
```

dcermak (approved):
Makes me a little sad that the performance tanked so hard with the new release, but we ain't got all time in the world…

PR #450, opened 2024-04-11
Add tests for the Spack container

e4t:
This is an initial implementation.

Unfortunately, a lot of the pytest_container 'goodness' was not usable to test the Spack container:

In order to test whether sending a Spack command directly to the container works, it needs to be launched:
 `podman run --rm <spack_container> <spack_command>`
 However, when running:
   container.connection.check_output()
a command will be sent to a via
 `podman exec <containerID> sh -c ...`
which will use not be able to run a Spack command.

Furthermore, we cannot use the MultiStageBuild.build() interface as the simplistic substitue() replacement will choke on '$..' shell constructs.

I'd suggest to merge this test into a separate branch and address issues there before merging  to main.


dirkmueller:
> In order to test whether sending a Spack command directly to the container works, it needs to be launched: `podman run --rm <spack_container> <spack_command>` However, when running: container.connection.check_output() a command will be sent to a via `podman exec <containerID> sh -c ...` which will use not be able to run a Spack command.

not a great way of doing it, but you can copy the method from here: 

https://github.com/SUSE/BCI-tests/blob/main/tests/test_helm.py#L12

e4t:
> > In order to test whether sending a Spack command directly to the container works, it needs to be launched: `podman run --rm <spack_container> <spack_command>` However, when running: container.connection.check_output() a command will be sent to a via `podman exec <containerID> sh -c ...` which will use not be able to run a Spack command.
> 
> not a great way of doing it, but you can copy the method from here:
> 
> https://github.com/SUSE/BCI-tests/blob/main/tests/test_helm.py#L12

not sure if you've seen the patch, but this was pretty much what I did:
https://github.com/e4t/BCI-tests/blob/238d877309efb46d044da5ec0e9cf9554deb0cc6/tests/test_spack.py#L60-L65
and:
https://github.com/e4t/BCI-tests/blob/238d877309efb46d044da5ec0e9cf9554deb0cc6/tests/test_spack.py#L99-L106
The difference being that this code uses `check_output()` while I use `run_expect()` which gives me a finer control over the container command return value: `check_output()` accepts 0 and 1.

tests/test_spack.py
```diff
@@ -0,0 +1,93 @@
+"""Tests for Spack application build container images."""
+# from pytest_container.build import MultiStageBuild
```
dcermak, line 2:
```suggestion
```

tests/test_spack.py
```diff
@@ -0,0 +1,93 @@
...
+    )
+
+    # run container with argument: 'containerize', save output to variable 'containerfile'
+    # Cannot use container.connection.run_expect(..,"containerize"):
+    # This uses `podman exec` which does not allow to test the `oneshot`
+    # feature of ENTRYPOINT: `podman run -v... --rm <container> containerize`
+    # containerfile = container.connection.run_expect(
+    #    [0],
+    #    "containerize"
+    # )
```
dcermak, line 67:
```suggestion
```

tests/test_spack.py
```diff
@@ -0,0 +1,93 @@
...
+    indirect=True,
+)
+def test_spack(
+    container,
+    host,
+    container_runtime,
+    tmp_path,
+    pytestconfig: Config,
+):
+    """Check if Spack Container allows to build an application container"""
```
dcermak, line 35: Please explain in more detail what you're doing here

tests/test_spack.py
```diff
@@ -0,0 +1,96 @@
...
+    Test if Spack Container can build a zsh container.
+
+    This function creates a `spack.yaml` input for spack, mounts it into the container,
+    runs the container with the 'containerize' argument which provides a `Containerfile`
+    multi-stage build description to build a zsh container.
+
+    For the final stage, the base container of the spack container is being used.
+
+    The test is building this description as a multi-stage container,
+    and finally tests whether the zsh in the resulting container is can be
```
dcermak, line 44:
```suggestion
    and finally tests whether the zsh in the resulting container can be
```

tests/test_spack.py
```diff
@@ -0,0 +1,96 @@
...
+from pytest_container import MultiStageBuild
+from pytest_container.container import BindMount
+from pytest_container.container import container_and_marks_from_pytest_param
+from pytest_container.container import DerivedContainer
+from pytest_container.helpers import get_extra_build_args
+from pytest_container.helpers import get_extra_run_args
+
+from bci_tester.data import BASE_CONTAINER
+from bci_tester.data import SPACK_CONTAINERS
+from bci_tester.runtime_choice import PODMAN_SELECTED
```
dcermak, line 15:
```suggestion
```

tests/test_spack.py
```diff
@@ -0,0 +1,96 @@
...
+
+# from pytest_container.container import Container
```
dcermak, line 17:
```suggestion
```

tests/test_spack.py
```diff
@@ -0,0 +1,96 @@
...
+
+CONTAINER_IMAGES = SPACK_CONTAINERS
+
+
+@pytest.mark.parametrize(
+    "container",
+    CONTAINER_IMAGES,
+    indirect=True,
+)
```
dcermak, line 26:
```suggestion
@pytest.mark.parametrize(
    "container",
    SPACK_CONTAINERS,
    indirect=True,
)
```

dcermak (changes requested):
Fails at the moment:
```
==> Error: `spack env activate` requires Spack's shell support.
  
  To set up shell support, run the command below for your shell.
  
  For bash/zsh/sh:
    . /usr/share/spack/setup-env.sh
  
  For csh/tcsh:
    source /usr/share/spack/setup-env.csh
  
  For fish:
    source /usr/share/spack/setup-env.fish
  
  For Windows batch:
    /usr/bin\spack_cmd.bat
  
  For PowerShell:
    /usr/share/spack\setup-env.ps1
  
  Or, if you do not want to use shell support, run one of these instead:
  
      eval `spack env activate --sh   [...]`  # bash/zsh/sh
      eval `spack env activate --csh  [...]`  # csh/tcsh
      eval `spack env activate --fish [...]`  # fish
      eval `spack env activate --bat  [...]`  # batch
      eval `spack env activate --pwsh [...]`  # powershell
  
  If you have already set up Spack's shell support but still receive
  this message, please make sure to call Spack via the `spack` command
  without any path components (such as `bin/spack`).
  
==> Error: install requires a package argument or active environment

Did you mean to install using the `spack.yaml` in this directory? Try: 
    spack env activate .
    spack install
  OR
    spack --env . install
```

dirkmueller:
> Fails at the moment:

this is fixed now (again)

dcermak (approved):

PR #452, opened 2024-04-11
Node test adjustments

dcermak:
- drop typescript, it takes ages to build, test and consumes insane amounts of memory
- resurect lodash test
- replace typescript with node-glob
- update docs

### DIFF
--- a/tests/test_node.py
+++ b/tests/test_node.py
@@ -30,10 +30,14 @@ def test_node_version(auto_container):
         for pkg in (
             GitRepositoryBuild(
                 repository_url="https://github.com/Microsoft/TypeScript",
-                build_command="npm ci && npm test",
-            )
-            if LOCALHOST.system_info.arch in ("x86_64",)
-            else None,
+                build_command="npm ci && npm run build",
+                marks=[
+                    pytest.mark.skipif(
+                        LOCALHOST.system_info.arch != "x86_64",
+                        reason="typescript only works reliably on x86_64",
+                    )
+                ],
+            ),
             GitRepositoryBuild(
                 repository_url="https://github.com/tj/commander.js.git",
                 build_command="npm ci && npm test && npm run lint",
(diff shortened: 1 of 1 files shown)


alexandrevicenzi (approved):

dirkmueller:
@dcermak the typescript tests seem to work after the revert to the previous podman version? you can see that in the action history there?

dcermak:
> @dcermak the typescript tests seem to work after the revert to the previous podman version? you can see that in the action history there?

They work, but they take ages, consume insane amounts of resources and I cannot execute them locally without either imposing memory limits or ooming my box (I guess there's the equivalent of `make -j $(nproc)` somewhere in the typescript test suite). Now they finish in a few minutes.

> typescript is the main framework that is used on top of node users, so covering it is kinda important..

Yes, that makes sense. Let me try to make typescript behave a bit.

dirkmueller:
We can run a subset of the testsuite also. I think we might want to set a cpulimit on the container or experiment with run-parallel-limit

dcermak:

> We can run a subset of the testsuite also. I think we might want to set a cpulimit on the container or experiment with run-parallel-limit

I have opted to just run `npm run build`, if I understand the
`package.json` correctly, then it should run some tests as well and it
finishes much quicker.

alexandrevicenzi (approved):

dirkmueller:
> I have opted to just run `npm run build`, if I understand the `package.json` correctly, then it should run some tests as well and it finishes much quicker.

its just building the tests and running the linter. thats not really test coverage. 

anyhow, I suggest to avoid the oom issue on your machine by limiting the worker count and only run light tests (which reduce the memory consumption), by e.g. executing

```
npm test -- --light --workers=4 
```

which finishes in about 1 minute 20s here, which I still find acceptable.

dcermak:
> ```
> npm test -- --light --workers=$(($(nproc)<4?$(nproc):4))
> ```
> 
> which finishes in about 1 minute 20s here, which I still find acceptable.

No idea how that finishes in such a short time on your machine. The whole test takes about 10 minutes on my box in both docker and podman.

dirkmueller:
> > which finishes in about 1 minute 20s here, which I still find acceptable.
> No idea how that finishes in such a short time on your machine. The whole test takes about 10 minutes on my box in both docker and podman.

I was just referring to the typescript part and I ran it outside of podman. in the pytest framework it is a bit slower:

```
181.36s call     tests/test_node.py::test_popular_npm_repos[bci/nodejs:20 from registry.opensuse.org/devel/bci/sle-15-sp5/containerfile/bci/nodejs:20-TypeScript]
```

dirkmueller (approved):
LGTM

PR #454, opened 2024-04-12
Increase the minimal image size per architecture

mgrossu:
VRs in openQA are failing for the bci-sp6 minimal image because of the size increase on all architectures: [aarch64](https://openqa.suse.de/tests/14011894#step/_root_BCI-tests_minimal_/1), [ppc64le](https://openqa.suse.de/tests/14011972#step/_root_BCI-tests_minimal_/1), [s390x](https://openqa.suse.de/tests/14011952), [x86_64](https://openqa.suse.de/tests/14014636#step/_root_BCI-tests_minimal_/1)
So, this PR bumps the size for the minimal image test.


dcermak:
This is a regression in rpm that should've been fixed

PR #459, opened 2024-04-22
Run test_base_size only on podman

grisu48:
Due to inconsistent space reporting between docker and podman we limit the image size test to podman only.

* Context: https://github.com/containers/podman/issues/13777#issuecomment-1090133955
* Related failures: https://openqa.suse.de/tests/14103942#step/_root_BCI-tests_base_/2

### DIFF
--- a/tests/test_base.py
+++ b/tests/test_base.py
@@ -19,6 +19,7 @@ from bci_tester.fips import FIPS_DIGESTS
 from bci_tester.fips import host_fips_enabled
 from bci_tester.fips import target_fips_enforced
 from bci_tester.runtime_choice import DOCKER_SELECTED
+from bci_tester.runtime_choice import PODMAN_SELECTED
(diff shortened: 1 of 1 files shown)


tests/test_base.py
```diff
@@ -34,6 +35,10 @@ def test_passwd_present(auto_container):
     assert auto_container.connection.file("/etc/passwd").exists
 
 
+@pytest.mark.skipif(
+    not PODMAN_SELECTED,
+    reason="inconsistent image size reporting between docker and podman",
```
dirkmueller, line 40: maybe something like `docker size reporting is dependant on underlying filesystem`

dirkmueller, line 40:
```suggestion
    reason="docker size reporting is dependant on underlying filesystem",
```

dirkmueller (approved):

grisu48:
Amended and rebased.

PR #461, opened 2024-04-23
Run tests using the chat command /run_tests toxenvs=mariadb,nginx

dcermak:
This should more or less allow us to run the BCI tests via a chatops command:

```
/run_tests toxevns=mariadb,nginx,openjdk,tomcat
```

I've made only the base set of tests run on each PR and the rest be triggered by a chat command. Unfortunately, we are now in a bit of an awkward spot: to test this, we need to merge this, as otherwise the `repository_dispatch` event is not sent & handled.

Also, this currently breaks our nightly CI: it will only run the base set of tests and not all of them

### DIFF
--- a/.github/workflows/command-dispatch.yml
+++ b/.github/workflows/command-dispatch.yml
@@ -1,0 +1,18 @@
+---
+name: Slash Command Dispatch
+on:
+  issue_comment:
+    types: [created]
+jobs:
+  dispatch-command:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Slash Command Dispatch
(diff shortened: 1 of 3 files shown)


dirkmueller:
> Also, this currently breaks our nightly CI: it will only run the base set of tests and not all of them

tbh for me that's unfortunately a no go. 

in my view, we should default to run all the tests, unless we identify its a PR that is touching only a set of / or maybe only a single file in tests/test_*.py. then those should be run. in all other cases, all tests should be run. 

having to write an extra command rather than just opening an PR and then leaving for coffee/lunch break is imho just too much effort. I want machines to work, not myself to work more :/

PR #464, opened 2024-04-25
Allow to restrict tox envs in test runs via PR description

dcermak:


.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -0,0 +1,8 @@
+<!--
+Thanks for sending a pull request!
+
+In case you are changing only tests for a specific environment and don't need to
+run all test environments, add the following line to your PR description on a
+separate line! [CI:TOXENVS] foo,bar
+
```
dirkmueller, line 7: maybe mention that `build,all,repository,metadata,multistage` is always executed in addition to what is specified

.github/workflows/ci.yaml
```diff
@@ -34,13 +33,31 @@ jobs:
...
+          set stringified_matrix (tox -l | sed -e '/unit/d' -e '/get_urls/d' -e '/doc/d' -e '/lint/d' -e '/fips/d' | jo -a)
+
+          set users_envs (echo $PR_BODY | awk -F' ' '/^\[CI:TOXENVS\]/ { print $2 }')
+          if [ -n "$users_envs" ]
+              set stringified_matrix (echo "$users_envs,build,all,repository,metadata,multistage" | tr ',' '\n' | sort | uniq | jo -a)
+          end
           echo "matrix=$stringified_matrix" >> $GITHUB_OUTPUT
+        shell: fish {0}
+        env:
+          PR_BODY: ${{ steps.pr.outputs.pr_body }}
```
dirkmueller, line 60: so this works when there isn't a PR ? it just resolves to empty then?

dirkmueller:
LGTM. I'd just do the determination of affected tests by looking at the list of modified files but this works as well.

alexandrevicenzi:
I don't see this working as expected. It is unlikely that external users will know what tests to run. Plus, you need to remember to edit the PR text before submitting it.

It should be the other way around perhaps. Run a few tests by default always, and an "admin" comments something to enable the rest of the tests to run.

Another issue is that tests all envs are required, if you run a subset only, the PR gets "blocked" by default.

dcermak:
> I don't see this working as expected. It is unlikely that external users will know what tests to run. 

We get zero external contributors. The majority of the PRs are by the BCI or the QA team who will know what to run.

> Plus, you need to remember to edit the PR text before submitting it.

That's why there's a template.

> It should be the other way around perhaps. Run a few tests by default always, and an "admin" comments something to enable the rest of the tests to run.

I did that in #461 and Dirk objected. Plus this solution is much simpler.

> 
> Another issue is that all env tests are required, if you run a subset only, the PR gets "blocked" by default.

Yes, but an admin can override that. Also, we can relax the required CI jobs.

.github/workflows/ci.yaml
dcermak: I have added a fallback and pushed the changes to my fork's main branch. Let's see what will happen tonight

.github/PULL_REQUEST_TEMPLATE.md
alexandrevicenzi: I noticed that in some of my PRs, GitHub interface grabs the last commit message and it would possibly hide this template.

Will this template show all the time? or can GitHub hide it in some scenarios?

dcermak:
Alexandre Vicenzi ***@***.***> writes:

> @alexandrevicenzi commented on this pull request.
>
>
>
>> @@ -0,0 +1,8 @@
> +<!--
> +Thanks for sending a pull request!
> +
> +In case you are changing only tests for a specific environment and don't need to
> +run all test environments, add the following line to your PR description on a
> +separate line! [CI:TOXENVS] foo,bar
> +
>
> I noticed that in some of my PRs, GitHub interface grabs the last commit message and it would possibly hide this template.
>
> Will this template show all the time? or can GitHub hide it in some scenarios?

Github can hide it sometimes, iirc if you create a PR explicitly from a
branch and only have one commit there.

dcermak:
@dirkmueller @alexandrevicenzi Any opinions how to move forward here?

dirkmueller (approved):

dirkmueller:
> @dirkmueller @alexandrevicenzi Any opinions how to move forward here?

see my earlier message already at https://github.com/SUSE/BCI-tests/pull/464#issuecomment-2084952391

PR #466, opened 2024-04-25
Run cassandra tests in separate workdir

dcermak:


dirkmueller:
I've done something similar in https://github.com/SUSE/BCI-tests/pull/442/files - it didn't help

PR #467, opened 2024-04-25
Split Base container size checks per version

dirkmueller:
This way we can more closely keep the margins in place for the pending fixes that will go into the various code streams over the next few weeks so that we can detect regressions earlier again rather than blanket having a really large size everywhere.


dcermak (approved):
LGTM, thanks!

PR #472, opened 2024-05-07
Add tests for GCC containers

dirkmueller:
won't do anything useful until https://github.com/SUSE/BCI-dockerfile-generator/pull/1088 has landed.


tests/test_gcc.py
```diff
@@ -0,0 +1,43 @@
...
+HELLO_CONTAINERS = [
+    pytest.param(
+        DerivedContainer(
+            base=container_and_marks_from_pytest_param(ctr)[0],
+            containerfile=CONTAINERFILE_HELLO,
+        ),
+        marks=ctr.marks,
+        id=ctr.id,
+    )
+    for ctr in CONTAINER_IMAGES
```
dcermak, line 32: `CONTAINER_IMAGES` is only necessary when you are using the `auto_container` fixture (that one parametrizes automatically).

```suggestion
_HELLO_VERSION = "2.12.1"

CONTAINERFILE_HELLO = f"""
WORKDIR /src
RUN curl -sLO https://ftpmirror.gnu.org/hello/hello-{_HELLO_VERSION}.tar.gz && \\
    tar --no-same-permissions --no-same-owner -xf hello-{_HELLO_VERSION}.tar.gz && \\
    cd hello-{_HELLO_VERSION} && \\
    ./configure && \\
    make && \\
    make install
"""

HELLO_CONTAINERS = [
    pytest.param(
        DerivedContainer(
            base=container_and_marks_from_pytest_param(ctr)[0],
            containerfile=CONTAINERFILE_HELLO,
        ),
        marks=ctr.marks,
        id=ctr.id,
    )
    for ctr in GCC_CONTAINERS
```

dcermak (changes requested):
Needs a rebase, otherwise LGTM

tests/test_gcc.py
dirkmueller: @dcermak we could use auto_container going forward? I find it useful to have the infrastructure for that in place already.

dcermak: There's nothing stopping you from using `auto_container` right now

dirkmueller: it doesn't have a use here.

dcermak (approved):

dcermak (changes requested):
Clicked too fast… Fails the metadata check, specifically supportlevel:
```
FAILED tests/test_metadata.py::test_l3_label[bci/gcc:7 from registry.opensuse.org/devel/bci/sle-15-sp5/containerfile/bci/gcc:7] - AssertionError: image supportlevel must be marked as L3
```

dirkmueller:
> Clicked too fast… Fails the metadata check, specifically supportlevel:

it needs https://github.com/SUSE/BCI-dockerfile-generator/pull/1110 to land first, which sets the containers to L3 as agreed with the toolchain team.

dcermak (approved):

PR #476, opened 2024-05-09
Test that iconv is working inside the base container

dirkmueller:
this requires a working installation of /usr/lib(64)?/gconv

### DIFF
--- a/tests/test_base.py
+++ b/tests/test_base.py
@@ -36,6 +36,20 @@ def test_passwd_present(auto_container):
 
 
 @pytest.mark.skipif(
+    OS_VERSION not in ("tumbleweed",),
+    reason="requires glibc-locale-base installed",
+)
+def test_iconv_working(auto_container):
+    """Generic test iconv works for UTF8 and ISO-8859-15 locale"""
+    assert (
+        auto_container.connection.check_output(
+            "echo -n 'SüSE' | iconv -f UTF8 -t ISO_8859-15 | wc -c"
+        )
+        == "4"
+    )
+
+
+@pytest.mark.skipif(
     not PODMAN_SELECTED,
     reason="docker size reporting is dependant on underlying filesystem",
 )


tests/test_base.py
```diff
@@ -35,6 +35,20 @@ def test_passwd_present(auto_container):
...
 
+@pytest.mark.skipif(
+    OS_VERSION not in ("tumbleweed",),
+    reason="requires glibc-locale-base installed",
+)
+def test_iconv_working(auto_container):
+    """Generic test iconv works for UTF8 and ISO-8859-15 locale"""
+    assert (
+        auto_container.connection.check_output(
+            "echo -n 'SüSE' | iconv -f UTF8 -t ISO_8859-15 | wc -c"
```
dcermak, line 46: Won't the pipe swallow the return code of iconv?

tests/test_base.py
```diff
@@ -35,6 +35,20 @@ def test_passwd_present(auto_container):
     assert auto_container.connection.file("/etc/passwd").exists
 
 
+@pytest.mark.skipif(
+    OS_VERSION not in ("tumbleweed",),
+    reason="requires glibc-locale-base installed",
```
dcermak, line 40: Wouldn't it make sense to install `glibc-locale-base` on non Tumbleweed?

dcermak (commented):
Is this testing for a specific regression/issue?

tests/test_base.py
dirkmueller: it does, but the wc -c is unlikely to be 4 in case of failure. we can also remove the wc -c to capture the exit code, not sure if that is better.

tests/test_base.py
dirkmueller: yes, it would totally make sense to ship base containers that are based on SLES components rather than kubic-locale-archive. but that's a change on dockerfile-generator..

dirkmueller:
> Is this testing for a specific regression/issue?

yes, I'm improving the packaging of glibc-locale-base and want to make sure that iconv continues to be working.

tests/test_base.py
dcermak: > but the wc -c is unlikely to be 4 in case of failure

Indeed, it'll probably be fine

PR #480, opened 2024-05-14
Test MARIADB_RANDOM_ROOT_PASSWORD

dirkmueller:


dcermak (commented):
I guess there's no way to easily check that a random password was actually used?

dirkmueller:
> I guess there's no way to easily check that a random password was actually used?

we could parse the start log and search for the password and connect as root to it. or we could simply test that we can not access without a password.

bci_tester/data.py
```diff
@@ -604,7 +604,7 @@ def create_BCI(
 PHP_8_APACHE = create_BCI(build_tag="bci/php-apache:8")
 PHP_8_FPM = create_BCI(build_tag="bci/php-fpm:8")
 
-MARIADB_ROOT_PASSWORD = "n0ts3cr3t"
+MARIADB_ROOT_PASSWORD = "'88tpw-n!t-s$$cr`t!"
```
dcermak, line 607: Jesus Christ, how did you drag that armadillo over your keyboard :see_no_evil:

tests/test_mariadb.py
```diff
@@ -84,14 +90,17 @@ def _wait_for_server(connection):
...
-    "container_per_test,db_user,db_password",
+    "container_per_test,db_user,db_password,root_pw",
     _generate_test_matrix(),
     indirect=["container_per_test"],
 )
 def test_mariadb_db_env_vars(
     container_per_test: ContainerData,
     db_user: str,
     db_password: str,
+    root_pw: str,
```
dcermak, line 101: Can be `None` too, see line 55
```suggestion
    root_pw: str | None,
```

tests/test_mariadb.py
```diff
@@ -135,6 +143,38 @@ def test_mariadb_db_env_vars(
...
+    if root_pw is None:
+        for line in host.check_output(
+            f"{container_runtime.runner_binary} logs {container_per_test.container_id}",
+        ).splitlines():
+            if "GENERATED ROOT PASSWORD: " in line:
+                root_pw = line.partition("GENERATED ROOT PASSWORD: ")[
+                    2
+                ].strip()
+                break
+
```
dcermak, line 155: The connect in the next line will certainly fail if it's `None`, but it'll make the error more readable

```suggestion
    assert root_pw, "Root password must be either set or obtained from the log"
```

dcermak (approved):

PR #482, opened 2024-05-15
Ignore boolean deps in the pubcloud maintenance release

dirkmueller:


dcermak (changes requested):
Fails for SP6 in `test_package_installation` with:
```
Package 'python311-azure-core' not found.
```

dirkmueller:
> Fails for SP6 in `test_package_installation` with:
> 
> ```
> Package 'python311-azure-core' not found.
> ```

that's just temporary, the same problem will be on SP6.

PR #483, opened 2024-05-16
Explicitly install libsolv-tools when we use it

dirkmueller:
libsolv-tools is no longer part of the default installed packages, so we need to pull it in for the test


PVlasin (commented):
OK

PR #490, opened 2024-05-30
Run FIPS tests in non-FIPS hosts

pablo-herranz:
For FIPS images we need to run the FIPS tests, even when those are running on non-FIPS hosts. Currently the FIPS tests are skipped on non-FIPS hosts. We need to add an exception for the FIPS images in BCI-Tests


dcermak (changes requested):
This is wrong, the current fips tests only work on hosts in FIPS mode. If you want to run a subset of them on a non-FIPS host, then you also need to modify the actual tests or add a new one that is not filtered out by the marker that you removed.

PR #491, opened 2024-05-30
Only replace the BCI_REPO url in containers where the file exists & is writeable

dcermak:
This is not a perfect solution as tomcat still works with the old url (it's build with `USER tomcat`), but it's better than before

### DIFF
--- a/bci_tester/data.py
+++ b/bci_tester/data.py
@@ -200,7 +200,10 @@ if BCI_DEVEL_REPO is None:
     BCI_DEVEL_REPO = f"https://updates.suse.com/SUSE/Products/SLE-BCI/{OS_MAJOR_VERSION}-SP{OS_SP_VERSION}/{LOCALHOST.system_info.arch}/product/"
     _BCI_REPLACE_REPO_CONTAINERFILE = ""
 else:
-    _BCI_REPLACE_REPO_CONTAINERFILE = f"RUN sed -i 's|baseurl.*|baseurl={BCI_DEVEL_REPO}|' /etc/zypp/repos.d/{BCI_REPO_NAME}.repo"
+    bci_repo_path = f"/etc/zypp/repos.d/{BCI_REPO_NAME}.repo"
+    # only try to replace the baseurl if the file exists and we have perl
+    # available (should be more ubiquitous than sed)
+    _BCI_REPLACE_REPO_CONTAINERFILE = f"RUN if [ -e {bci_repo_path} ] && [ -e /usr/bin/perl ]; then perl -pi -e 's|baseurl.*|baseurl={BCI_DEVEL_REPO}|g' {bci_repo_path}; fi"
 
 assert BCI_DEVEL_REPO, "BCI_DEVEL_REPO must be set at this point"
 
@@ -250,6 +253,7 @@ def create_BCI(
     available_versions: Optional[List[str]] = None,
     extra_marks: Optional[Sequence[MarkDecorator]] = None,
     bci_type: ImageType = ImageType.LANGUAGE_STACK,
+    container_user: Optional[str] = None,
     **kwargs,
 ) -> ParameterSet:
     """Creates a DerivedContainer wrapped in a pytest.param for the BCI with the
@@ -333,12 +337,18 @@ def create_BCI(
     else:
         baseurl = f"{BASEURL}/{_get_repository_name(image_type)}{build_tag}"
 
+    if bci_type == ImageType.OS_LTSS:
+        containerfile = ""
+    else:
+        if container_user:
+            containerfile = f"USER root\n{_BCI_REPLACE_REPO_CONTAINERFILE}\nUSER {container_user}"
+        else:
+            containerfile = _BCI_REPLACE_REPO_CONTAINERFILE
+
     return pytest.param(
         DerivedContainer(
             base=baseurl,
-            containerfile=""
-            if bci_type == ImageType.OS_LTSS
-            else _BCI_REPLACE_REPO_CONTAINERFILE,
+            containerfile=containerfile,
             **kwargs,
         ),
         marks=marks,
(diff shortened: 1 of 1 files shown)


bci_tester/data.py
```diff
@@ -199,8 +199,11 @@ def create_container_version_mark(
     BCI_DEVEL_REPO = f"https://updates.suse.com/SUSE/Products/SLE-BCI/{OS_MAJOR_VERSION}-SP{OS_SP_VERSION}/{LOCALHOST.system_info.arch}/product/"
     _BCI_REPLACE_REPO_CONTAINERFILE = ""
 else:
-    _BCI_REPLACE_REPO_CONTAINERFILE = f"RUN sed -i 's|baseurl.*|baseurl={BCI_DEVEL_REPO}|' /etc/zypp/repos.d/{BCI_REPO_NAME}.repo"
+    bci_repo_path = f"/etc/zypp/repos.d/{BCI_REPO_NAME}.repo"
+    # only try to sed the baseurl if we can write to the repo
+    _BCI_REPLACE_REPO_CONTAINERFILE = f"RUN if [ -w {bci_repo_path} ]; then sed -i 's|baseurl.*|baseurl={BCI_DEVEL_REPO}|' {bci_repo_path}; fi"
 
+assert BCI_DEVEL_REPO, "BCI_DEVEL_REPO must be set at this point"
```
dirkmueller, line 206: I don't understand this assert. isn't that always wrong for LTSS containers?

dcermak: The variable is set to a value in line 199 if it is set to `None`.

bci_tester/data.py
```diff
@@ -325,9 +327,11 @@ def create_BCI(
     return pytest.param(
         DerivedContainer(
             base=baseurl,
-            containerfile=""
-            if bci_type == ImageType.OS_LTSS
-            else _BCI_REPLACE_REPO_CONTAINERFILE,
+            containerfile=(
```
dirkmueller, line 331: is this a automatic code formatter change?

bci_tester/data.py
dirkmueller: yeah, thanks. a 2nd look helped. sorry for the false noise.

bci_tester/data.py
```diff
@@ -199,8 +199,11 @@ def create_container_version_mark(
     BCI_DEVEL_REPO = f"https://updates.suse.com/SUSE/Products/SLE-BCI/{OS_MAJOR_VERSION}-SP{OS_SP_VERSION}/{LOCALHOST.system_info.arch}/product/"
     _BCI_REPLACE_REPO_CONTAINERFILE = ""
 else:
-    _BCI_REPLACE_REPO_CONTAINERFILE = f"RUN sed -i 's|baseurl.*|baseurl={BCI_DEVEL_REPO}|' /etc/zypp/repos.d/{BCI_REPO_NAME}.repo"
+    bci_repo_path = f"/etc/zypp/repos.d/{BCI_REPO_NAME}.repo"
+    # only try to sed the baseurl if we can write to the repo
+    _BCI_REPLACE_REPO_CONTAINERFILE = f"RUN if [ -w {bci_repo_path} ]; then sed -i 's|baseurl.*|baseurl={BCI_DEVEL_REPO}|' {bci_repo_path}; fi"
```
dirkmueller, line 204: Why don't we do something like

```
USER root
RUN sed -i 's|baseurl.*|baseurl={BCI_DEVEL_REPO}|' /etc/zypp/repos.d/{BCI_REPO_NAME}.repo
USER {containeruser}
```

instead? sure it requires us to specify the containeruser as part of bci_tester data but that has the advantage that we can check for the container actually running as non root (and as that user) which might be an important security property worth checking for.

bci_tester/data.py
dcermak: yes, I forced it to be more readable by adding the braces

bci_tester/data.py
dcermak: I was thinking about doing this, but I'm not a huge fan of this approach. Afaik we only have one container with a different user (tomcat), so this is special handling because of one container and more importantly, we don't test this code path in the github CI at all. So if we mess up anything in here, we'll blow up openQA without noticing.

dirkmueller: we have more than one, distribution/registry have a different user. we can test this codepath in the github ci very easily by setting the environment variable, so that's not a good reason. 

also dotnet containers are non-root meanwhile. so we need to have a generic solution. it's not acceptable to *not* change the repository as it invalidates the testng results (we're not testing with the repo under test)

dcermak: I've implemented that now

bci_tester/data.py
```diff
@@ -200,8 +200,12 @@ def create_container_version_mark(
     BCI_DEVEL_REPO = f"https://updates.suse.com/SUSE/Products/SLE-BCI/{OS_MAJOR_VERSION}-SP{OS_SP_VERSION}/{LOCALHOST.system_info.arch}/product/"
     _BCI_REPLACE_REPO_CONTAINERFILE = ""
 else:
-    _BCI_REPLACE_REPO_CONTAINERFILE = f"RUN sed -i 's|baseurl.*|baseurl={BCI_DEVEL_REPO}|' /etc/zypp/repos.d/{BCI_REPO_NAME}.repo"
+    bci_repo_path = f"/etc/zypp/repos.d/{BCI_REPO_NAME}.repo"
+    # only try to replace the baseurl if the file exists and we have perl
+    # available (should be more ubiquitous than sed)
```
dirkmueller, line 205: We have more images with perl than with sed?

bci_tester/data.py
```diff
@@ -200,8 +200,12 @@ def create_container_version_mark(
...
+    _BCI_REPLACE_REPO_CONTAINERFILE = f"RUN if [ -e {bci_repo_path} ] && [ -e /usr/bin/perl ]; then perl -pi -e 's|baseurl.*|baseurl={BCI_DEVEL_REPO}|g' {bci_repo_path}; fi"
```
dirkmueller, line 206: So this switches from unconditional replacement to conditional execution. Why?

bci_tester/data.py
dcermak: perl is a hard dependency of rpm and thereby guaranteed to be in every container except minimal & busybox. It doesn't really matter in practice at the moment (as everything without sed also lacks zypper and thereby these tests get skipped). But it should be more future proof.

bci_tester/data.py
dcermak: Because it makes no sense to modify a non-existing file with a binary that doesn't exist?

PR #497, opened 2024-06-05
Mark SP6 based containers as released

dirkmueller:
Also sorts the lists alphabetically to make it less confusing. Spack containers had to be added to the L3 containers list


dcermak (changes requested):
- Can you make this two commits please? The actually important change is completely lost.
- https://build.opensuse.org/projects/devel:BCI:SLE-15-SP6/packages/sles15-image/files/config.kiwi?expand=1 is still marked as techpreview, so the metadata test needs an xfail

dirkmueller:
There's a pending request to fix bci-base

PR #499, opened 2024-06-05
Restore sizing for fips container

dirkmueller:


tests/test_base.py
```diff
@@ -69,7 +69,11 @@ def test_base_size(auto_container: ContainerData, container_runtime):
     )
 
     #: size limits of the base container per arch in MiB
-    if OS_VERSION in ("basalt", "tumbleweed") or is_fips_ctr:
+    if is_fips_ctr:
+        base_container_max_size: Dict[str, int] = {
+            "x86_64": 130,
+        }
```
dcermak, line 75: Don't we ship the container for non-x86_64?

dirkmueller: nope, fips is x86_64 only

dcermak (approved):

msmeissn:
(FWIW we certify FIPS for x86_64, s390x and aarch64 currently. Just we only have a x86_64 fips container)

PR #502, opened 2024-06-10
Upgrade Helm to a version that doesn't have failing tests

dirkmueller:
Unfortunately this requires disabling tests on go < 1.22 because the new version requires it. this is only temporary so I think we can live with that.

### DIFF
--- a/tests/test_go.py
+++ b/tests/test_go.py
@@ -44,7 +44,7 @@ def test_go_version(auto_container):
         GitRepositoryBuild(
             repository_url="https://github.com/weaveworks/kured.git",
             repository_tag="1.13.2",
-            build_command="go run ./cmd/kured -h && go test -race ./...",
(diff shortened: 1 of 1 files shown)


tests/test_go.py
```diff
@@ -79,6 +79,13 @@ def test_build_helm(auto_container_per_test, container_git_clone):
     container with :command:`make` pre-installed.
 
     """
+
+    if (
+        "oldstable" in auto_container_per_test.image_url_or_id
+        or "stable-openssl" in auto_container_per_test.image_url_or_id
+    ):
+        pytest.skip("Helm requires Go >= 1.22")
```
dcermak, line 87: Wouldn't it make more sense to explicitly check the go version? Then we won't accidentally skip images in the future in case we forget

dirkmueller: can do as well

PR #513, opened 2024-06-19
Install libsolv-tools for the installcheck

dirkmueller:
15.6 and newer has gotten the new libzypp stack which strips libsolv-tools as a default dependency. so we need to install it explicitly as it is no longer in the container by default.


dcermak (commented):
LGTM, but needs https://build.suse.de/request/show/336299 to work first

dirkmueller:
> LGTM, but needs https://build.suse.de/request/show/336299 to work first

I've seen that already. we also need it for SP5.

dirkmueller:
publishing of bci repo fails because the tests are failing without this fix.

PR #514, opened 2024-06-21
Install procps for cassandra tests

dcermak:
One of the scripts calls free that is included in the procps package and no longer shipped by default

[CI:TOXENVS] openjdk


dirkmueller:
still needs  `installcheck:` command not found`

dirkmueller (approved):

PR #517, opened 2024-06-24
Enable ruff isort

dcermak:
I merged the ruff PR too quickly without checking that isort is enabled


dirkmueller:
It's sort of surprising how a pr that was not marked as ready can be merged without checking back with the author, given that it has a clear todo on the comments..

dirkmueller (approved):

PR #520, opened 2024-06-25
Switch 15.6

dirkmueller:
accidentally merged #519 too early, restore to see CI runs


dirkmueller:
tests look good, with failures only due to bci repo issue and was reviewed before already.

PR #522, opened 2024-06-27
Test for codestream lifecycle information being correct (bsc#1227100)

dirkmueller:
[CI:TOXENVS] repository

### DIFF
--- a/tests/test_repository.py
+++ b/tests/test_repository.py
@@ -181,6 +181,31 @@ def test_repo_content_licensing(container_per_test) -> None:
     OS_VERSION not in ALLOWED_BCI_REPO_OS_VERSIONS,
     reason="no included BCI repository - can't test",
 )
+@pytest.mark.parametrize("container_per_test", [BASE_CONTAINER], indirect=True)
+def test_codestream_lifecycle(container_per_test):
+    """Check that the codestream lifecycle information is available
+    and has the expected value."""
+
+    zypper_lifecycle_xml = ET.fromstring(
+        container_per_test.connection.check_output(
+            "zypper --xmlout -i pd --xmlfwd codestream"
+        )
+    )
+    lifecycle = zypper_lifecycle_xml.find(
+        ".//product[@name='SLES']/xmlfwd/codestream/endoflife"
+    )
+    assert (
+        lifecycle is not None
+    ), "No endoflife information found in product description"
+    assert (
+        lifecycle.text == "2031-07-31"
+    ), f"Expected end of life 2031-07-31, but got {lifecycle.text}"
+
+
+@pytest.mark.skipif(
+    OS_VERSION not in ALLOWED_BCI_REPO_OS_VERSIONS,
+    reason="no included BCI repository - can't test",
+)
 @pytest.mark.skipif(
     OS_VERSION == "tumbleweed", reason="No testing for openSUSE"
 )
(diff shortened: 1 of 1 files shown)


dcermak (commented):
This still needs a reformat and also the repo currently shows 2028 as the EOL date, so I guess this needs to be fixed in SLES?

dirkmueller:
> This still needs a reformat and also the repo currently shows 2028 as the EOL date, so I guess this needs to be fixed in SLES?

Yeah, I'm following TDD

dcermak:
What is blocking getting this fixed on the SLES side? Something in the SLE_BCI repo that we can fix?

dirkmueller:
> What is blocking getting this fixed on the SLES side? Something in the SLE_BCI repo that we can fix?

the information comes from the sles-release package. since it was wrong in SP6:GA, we need to release a maintenance update that noone has submitted so far.

tests/test_repository.py
```diff
@@ -157,6 +157,31 @@ def test_package_installation(container_per_test, pkg):
     )
 
 
+@pytest.mark.skipif(
+    OS_VERSION not in ALLOWED_BCI_REPO_OS_VERSIONS,
+    reason="no included BCI repository - can't test",
```
alexandrevicenzi, line 162: > not included in the BCI repository

this is a skipif, there's no need to include "can't test", the logs will already produce a message like:

> SKIPPED: not included in the BCI repository

dirkmueller: that changes the meaning.. (it's not about it not being in the bci repo but about the bci repo not being in those containers)

alexandrevicenzi: still not clear to me, but ok.

dirkmueller: I've removed it everywhere now (it was just a copy of all the previous messages)

alexandrevicenzi (approved):

tests/test_repository.py
```diff
@@ -159,7 +159,32 @@ def test_package_installation(container_per_test, pkg):
 
 @pytest.mark.skipif(
     OS_VERSION not in ALLOWED_BCI_REPO_OS_VERSIONS,
-    reason="no included BCI repository - can't test",
+    reason="no included BCI repository",
+)
```
dcermak, line 163: This needs a skipif for Tumbleweed too

dcermak (changes requested):

tests/test_repository.py
dirkmueller: also for 16.0. I made it 15.x specific.

tests/test_repository.py
```diff
@@ -177,9 +177,37 @@ def test_repo_content_licensing(container_per_test) -> None:
     ), "Found a package with a SUSE-Firmware license"
 
 
+@pytest.mark.skipif(
+    (
+        OS_VERSION.startswith("15")
+        and OS_VERSION not in ALLOWED_BCI_REPO_OS_VERSIONS
+    ),
+    reason="no included BCI repository",
+)
```
dcermak, line 186: Are you sure this is the right condition? Because the test now runs on Tumbleweed

dirkmueller: now it is.

alexandrevicenzi (approved):

dcermak (approved):

PR #529, opened 2024-07-09
Marks for all os versions

dcermak:
[CI:TOXENVS] base

### DIFF
--- a/bci_tester/data.py
+++ b/bci_tester/data.py
@@ -305,9 +305,7 @@ def create_BCI(
     # only try to grab the mark from the build tag for containers that are
     # available for this os version, otherwise we get bogus errors for missing
     # marks
-    if OS_VERSION in (
-        available_versions or list(_DEFAULT_NONBASE_OS_VERSIONS)
-    ):
+    if OS_VERSION in (available_versions or list(ALLOWED_NONBASE_OS_VERSIONS)):
         marks.append(pytest.mark.__getattr__(build_tag_base.replace(":", "_")))
 
     if OS_VERSION == "tumbleweed":
--- a/bci_tester/data.py
+++ b/bci_tester/data.py
@@ -345,6 +345,7 @@ else:
         build_tag=f"{BCI_CONTAINER_PREFIX}/bci-base:{OS_CONTAINER_TAG}",
         image_type="kiwi",
         bci_type=ImageType.OS,
+        available_versions=[OS_VERSION],
     )
     if TARGET in ("ibs", "ibs-cr", "ibs-released"):
         LTSS_BASE_CONTAINERS.extend(


bci_tester/data.py
```diff
@@ -347,6 +345,7 @@ def create_BCI(
         build_tag=f"{BCI_CONTAINER_PREFIX}/bci-base:{OS_CONTAINER_TAG}",
         image_type="kiwi",
         bci_type=ImageType.OS,
+        available_versions=[OS_VERSION],
```
dirkmueller, line 348: why not `_DEFAULT_NONBASE_OS_VERSIONS` ?

bci_tester/data.py
```diff
@@ -305,9 +305,7 @@ def create_BCI(
     # only try to grab the mark from the build tag for containers that are
```
dirkmueller, line 305: so this comment is obsolete now? what made it obsolete?

bci_tester/data.py
dirkmueller: why is this not needed for the tumbleweed case above?

dcermak: Because that's wrong, it includes `15.6` and `tumbleweed`, which is most certainly not the os_version for which `bci-base:15.5` is available.

bci_tester/data.py
dcermak: it is still correct, e.g. for 15.4 & 15.3. We don't have markers in the pyproject.toml for the sp3 & sp4 language containers (that don't exist anymore)

bci_tester/data.py
dirkmueller: it seems to me the issue is that available_versions for base  (so not nonbase) containers is not expanded properly. 

everything appears to be working with just this change, which would be logical to me:

```patch
--- a/bci_tester/data.py
+++ b/bci_tester/data.py
@@ -53,6 +53,9 @@ _DEFAULT_NONBASE_SLE_VERSIONS = ("15.5", "15.6")
 # Test Language and Application containers by default for these versions
 _DEFAULT_NONBASE_OS_VERSIONS = ("15.6", "tumbleweed")
 
+# Test base containers by default for these versions
+_DEFAULT_BASE_OS_VERSIONS = ("15.5", "15.6", "tumbleweed")
+
 assert (
     sorted(ALLOWED_BASE_OS_VERSIONS) == list(ALLOWED_BASE_OS_VERSIONS)
 ), f"list ALLOWED_BASE_OS_VERSIONS must be sorted, but got {ALLOWED_BASE_OS_VERSIONS}"
@@ -298,6 +301,15 @@ def create_BCI(
                     )
         else:
             available_versions = list(_DEFAULT_NONBASE_OS_VERSIONS)
+    else:
+        if available_versions:
+            for ver in available_versions:
+                if ver not in ALLOWED_BASE_OS_VERSIONS:
+                    raise ValueError(
+                        f"Invalid os version for base container: {ver}"
+                    )
+        else:
+            available_versions = list(_DEFAULT_BASE_OS_VERSIONS)
 
     if available_versions:
```

bci_tester/data.py
dirkmueller: ok, got it, so this if condition is only for nonbase containers.

bci_tester/data.py
dirkmueller: submitted now as alternative here: https://github.com/SUSE/BCI-tests/pull/531

PR #531, opened 2024-07-10
Restore base container marks for 15.5

dirkmueller:
The automatic setting of default marks only was working for development and application containers. restore it for base containers, as right now 15.5 is not in dev/app but it is in base containers.

### DIFF
--- a/bci_tester/data.py
+++ b/bci_tester/data.py
@@ -306,7 +306,7 @@ def create_BCI(
             for ver in available_versions:
                 if ver not in ALLOWED_BASE_OS_VERSIONS:
                     raise ValueError(
-                        f"Invalid os version for base container: {ver}"
+                        f"Invalid os version {ver} for base container: {build_tag}"
                     )
         else:
             available_versions = list(_DEFAULT_BASE_OS_VERSIONS)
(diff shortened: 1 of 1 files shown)


mgrossu (approved):

bci_tester/data.py
```diff
@@ -298,6 +301,15 @@ def create_BCI(
...
             available_versions = list(_DEFAULT_NONBASE_OS_VERSIONS)
+    else:
+        if available_versions:
+            for ver in available_versions:
+                if ver not in ALLOWED_BASE_OS_VERSIONS:
+                    raise ValueError(
+                        f"Invalid os version for base container: {ver}"
+                    )
+        else:
+            available_versions = list(_DEFAULT_BASE_OS_VERSIONS)
```
dcermak, line 312: This does not make a lot of sense for os & LTSS containers, as these are really only valid for a **single** os version. We will probably never hit a bug due to that, as we usually set their tags from `OS_VERSION`, but imho we could also just do this:

```Python
    else:
        if not available_versions:
            raise ValueError("OS & LTSS containers must have a version marker")
        else:
            for ver in available_versions:
                if ver not in ALLOWED_BASE_OS_VERSIONS:
                    raise ValueError(
                        f"Invalid os version for base container: {ver}"
                    )
```

bci_tester/data.py
```diff
@@ -298,6 +301,15 @@ def create_BCI(
                     )
         else:
             available_versions = list(_DEFAULT_NONBASE_OS_VERSIONS)
+    else:
+        if available_versions:
+            for ver in available_versions:
+                if ver not in ALLOWED_BASE_OS_VERSIONS:
+                    raise ValueError(
+                        f"Invalid os version for base container: {ver}"
```
dcermak, line 309:
```suggestion
                        f"Invalid os version {ver} for base container: {build_tag}"
```

PR #532, opened 2024-07-11
Reduce tumbleweed size (it's ~ 99 MB now)

dirkmueller:

### DIFF
--- a/tests/test_base.py
+++ b/tests/test_base.py
@@ -37,7 +37,7 @@ def test_passwd_present(auto_container):
 
 @pytest.mark.skipif(
     OS_VERSION not in ("tumbleweed",),
-    reason="requires glibc-locale-base installed",
+    reason="requires gconv modules",
 )
 def test_iconv_working(auto_container):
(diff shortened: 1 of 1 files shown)


tests/test_base.py
```diff
@@ -37,7 +37,7 @@ def test_passwd_present(auto_container):
 
 @pytest.mark.skipif(
     OS_VERSION not in ("tumbleweed",),
-    reason="requires glibc-locale-base installed",
+    reason="requires gconv modules installed",
```
dcermak, line 40:
```suggestion
    reason="requires gconv modules to be installed",
```

dcermak (approved):
LGTM, but maybe rebase to be safe?

alexandrevicenzi (dismissed):

tests/test_base.py
dirkmueller: I wanted shorter, not longer. I've removed installed now.

dirkmueller:
> LGTM, but maybe rebase to be safe?

it has two approvals already...

PR #534, opened 2024-07-16
Add TW spack testing

dirkmueller:

### DIFF
--- a/bci_tester/data.py
+++ b/bci_tester/data.py
@@ -744,7 +744,7 @@ SPACK_CONTAINERS = [
         build_tag=f"{BCI_CONTAINER_PREFIX}/spack:{tag}",
         available_versions=[f"{ver}"],
     )
-    for ver, tag in (("15.6", "0.21"),)
+    for ver, tag in (("15.6", "0.21"), ("tumbleweed", "0.22"))
 ]
 
 PROMETHEUS_CONTAINERS = [
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -110,6 +110,7 @@ markers = [
     'bci-sle15-kernel-module-devel_15.5',
     'bci-sle15-kernel-module-devel_15.6',
     'spack_0.21',
+    'spack_0.22',
 ]
 
 [tool.ruff]


dcermak:
This fails only with podman (and not with docker) with:
```
----------------------------- Captured stderr call -----------------------------
/usr/bin/tar: pcre-8.45/cmake: Cannot change mode to rwxr-xr-x: Operation not permitted
/usr/bin/tar: pcre-8.45/m4: Cannot change mode to rwxr-xr-x: Operation not permitted
/usr/bin/tar: pcre-8.45/sljit: Cannot change mode to rwxr-xr-x: Operation not permitted
/usr/bin/tar: pcre-8.45/testdata: Cannot change mode to rwxr-xr-x: Operation not permitted
/usr/bin/tar: pcre-8.45/doc/html: Cannot change mode to rwxr-xr-x: Operation not permitted
/usr/bin/tar: pcre-8.45/doc: Cannot change mode to rwxr-xr-x: Operation not permitted
/usr/bin/tar: pcre-8.45: Cannot change mode to rwxr-xr-x: Operation not permitted
/usr/bin/tar: Exiting with failure status due to previous errors
==> Error: ProcessError: Command exited with status 2:
    '/usr/bin/tar' '-oxf' '/tmp/root/spack-stage/spack-stage-pcre-8.45-skm3fvdkmhaayc5hgrucm4okz24qkbry/pcre-8.45.tar.bz2'
==> Warning: Skipping build of zsh-5.8-ldow7yzmcdkp2egkncjpfbqxzbjl2low since pcre-8.45-skm3fvdkmhaayc5hgrucm4okz24qkbry failed
==> Error: Terminating after first install failure: ProcessError: Command exited with status 2:
    '/usr/bin/tar' '-oxf' '/tmp/root/spack-stage/spack-stage-pcre-8.45-skm3fvdkmhaayc5hgrucm4okz24qkbry/pcre-8.45.tar.bz2'
error building at STEP "RUN cd /opt/spack-environment && spack env activate . && spack install --fail-fast && spack gc -y": error while running runtime: exit status 1
```

Iirc you already fixed a comparable issue in the past Dirk, do you recall what the fix was?

dirkmueller:
> This fails only with podman (and not with docker) with:

> Iirc you already fixed a comparable issue in the past Dirk, do you recall what the fix was?

we don't have a fix. the problem is that tumbleweed uses some syscalls that are blocked on the workers and fails. the workaround was to add `--no-same-permissions --no-same-owner`

PR #542, opened 2024-07-19
Fix the url to download the GNU hello test project

dcermak:
[CI:TOXENVS] gcc

### DIFF
--- a/tests/test_gcc.py
+++ b/tests/test_gcc.py
@@ -13,7 +13,7 @@ _HELLO_VERSION = "2.12.1"
 
 CONTAINERFILE_HELLO = f"""
 WORKDIR /src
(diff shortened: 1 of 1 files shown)


tests/test_gcc.py
```diff
@@ -13,7 +13,7 @@
 
 CONTAINERFILE_HELLO = f"""
 WORKDIR /src
-RUN curl -sLO https://ftpmirror.gnu.org/hello/hello-{_HELLO_VERSION}.tar.gz && \\
+RUN curl -sLO https://ftp.gnu.org/gnu/hello/hello-{_HELLO_VERSION}.tar.gz && \\
```
dirkmueller, line 16: you're just changing the mirror redirector to a hardcoded site. the webpage explicitly recommends against this practice. 

the redirector is some times a bit flakey, it will recover. the change is imho not needed.

PR #543, opened 2024-07-19
Rewrite test_disturl_can_be_checked_out to not rely on osc

dirkmueller:
The test either does not work because osc is not installed, or it fails because osc is available but its dependency are not available inside the venv that the testsuite is running under. We can easily rewrite that with just using requests and avoiding the conflict with venvs

### DIFF
--- a/tests/test_metadata.py
+++ b/tests/test_metadata.py
@@ -14,10 +14,12 @@ offer the labels under the ``com.suse.bci`` prefix but ``com.suse.sle``.
 
 """
 
-from subprocess import check_output
+import urllib.parse
+from pathlib import Path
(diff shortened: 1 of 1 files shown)


dcermak (commented):
We will have to skip this new test for the `ibs-released` target in github actions, as they cannot resolve `build.suse.de`

dcermak (dismissed):

dcermak (commented):
Looks like this is failing for Tumbleweed

dirkmueller:
> Looks like this is failing for Tumbleweed

that's fixed now.

dcermak (approved):

PR #544, opened 2024-07-22
Stop testing on SP5

dirkmueller:
We have switched over all non-base containers to SP6. Testing the SP5 containers no longer makes sense.


dcermak (changes requested):
We also need to keep testing the language stack & app-containers that are still build from sp5 (e.g. postgres 15, java 11)

dirkmueller:
right, then I think this isn't worth it right now.

PR #546, opened 2024-07-25
Enable testing of "base" container on 15 SP7

dirkmueller:


tests/test_base.py
```diff
@@ -215,6 +222,10 @@ def test_all_openssl_hashes_known(auto_container):
     not DOCKER_SELECTED,
     reason="Docker in docker can only be tested when using the docker runtime",
 )
+@pytest.mark.skipif(
+    OS_VERSION in ("15.7",),
+    reason="SLE BCI repo not yet available",
+)
```
dcermak, line 228: maybe xfail this instead so that we don't forget about the test?

dirkmueller: yep, good point, I forgot about that.

alexandrevicenzi (approved):

PR #548, opened 2024-07-25
Initial enablement for the base-fips container

dirkmueller:


alexandrevicenzi (approved):

bci_tester/data.py
```diff
@@ -365,6 +365,7 @@ def create_BCI(
     for ver, tag in (("15.6", "9.24"), ("tumbleweed", "10.0"))
 ]
 
+BASE_FIPS_CONTAINERS = []
```
dcermak, line 368: Does it make things simpler if we switched this to a single constant?

dcermak (changes requested):
LGTM, except that we have to skip the base-fips image for sp7

bci_tester/data.py
dirkmueller: you mean not a list but a single parameterset? it was that originally, but there are OS_VERSIONS for which none exists so it's easier to handle the "test this container if it exists otherwise do nothing" approach by having a list with a single entry or an empty list. less code duplication imho.

dirkmueller:
> LGTM, except that we have to skip the base-fips image for sp7

It is skipped now. Also rebased to fix the merge conflict with #553

dcermak (approved):
LGTM, thanks!

Ignore the TW failure, it's due to a libzypp upgrade

PR #551, opened 2024-07-30
enable `test_kiwi_create_image` for podman container runtime

Priyankasaggu11929:
PR enables `test_kiwi_create_image()` for podman container runtime as well.
(earlier enabled only for `docker` assuming podman rootful mode wasn't allowed)
    
Verified – GitHub CI tests runs as root for podman
ref: https://github.com/SUSE/BCI-tests/blob/2de28a851adfd596fae5213c62472e7d01f0e043/.github/workflows/ci.yaml#L356-L357

---

Verified in CI, the kiwi tests executes successfully!

- Tumbleweed – https://github.com/SUSE/BCI-tests/actions/runs/10160098349/job/28095780381?pr=551#step:13:38
- 15.6 – https://github.com/SUSE/BCI-tests/actions/runs/10160098349/job/28095779927?pr=551#step:13:38

### DIFF
--- a/tests/test_kiwi.py
+++ b/tests/test_kiwi.py
@@ -7,7 +7,7 @@ from pytest_container.container import ContainerData
 from pytest_container.container import ImageFormat
 
 from bci_tester.data import KIWI_CONTAINERS
-from bci_tester.runtime_choice import DOCKER_SELECTED
+from bci_tester.runtime_choice import PODMAN_SELECTED
 
 CONTAINER_IMAGES = KIWI_CONTAINERS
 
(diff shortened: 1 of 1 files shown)


Priyankasaggu11929:
cc: @dirkmueller for review. Thanks!

alexandrevicenzi (approved):

tests/test_kiwi.py
```diff
@@ -50,17 +49,6 @@ def test_kiwi_installation(auto_container):
...
 
-# Disabling this test with `podman` as it mounts the newly created iso image (by kiwi)
-# on a `/dev/loop*` device, which needs a rootful container for `root:disk` ownership.
-# In rootless mode, `/dev/loop*` devices are owned by `nobody:nobody`,
-# causing `losetup -f --show /tmp/myimage/kiwi-test-image-disk.x86_64-1.15.3.raw` to fail with "Permission denied".
-#
-# Also ref: https://github.com/containers/podman/issues/17715#issuecomment-1460227771
-# Mounting a loop device in rootless mode is not allowed by the kernel.
-@pytest.mark.skipif(
-    not DOCKER_SELECTED,
```
dirkmueller, line 61: thanks, but this still breaks for non-root podman. can we add a skipif for the case of nonroot podman? maybe the assert on the loop device permission needs to be turned into a skip to achieve that?

Priyankasaggu11929: @dirkmueller, added a skipif for non-root podman case.
 
 Question - [here ](https://github.com/SUSE/BCI-tests/blob/main/.github/workflows/ci.yaml#L358-L364) is the only place where I understood tests are run with podman (in this case, with `sudo`). How/Where are the rootless podman tests triggered?

dirkmueller: not in CI, but in manual execution (most BCI developers use rootless podman and run tests locally)

dirkmueller (approved):

PR #555, opened 2024-08-02
Fix handling of the ironbank base container

dirkmueller:
The ironbank base container is a fips container but actually called bci-base.

### DIFF
--- a/bci_tester/data.py
+++ b/bci_tester/data.py
@@ -379,15 +379,16 @@ else:
         image_type="kiwi",
         bci_type=ImageType.OS,
     )
-    BASE_FIPS_CONTAINERS = [
-        create_BCI(
-            build_tag=f"{BCI_CONTAINER_PREFIX}/bci-base-fips:{OS_CONTAINER_TAG}",
-            bci_type=ImageType.OS,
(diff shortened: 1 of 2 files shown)


dcermak (commented):
Would it maybe make the handling a bit simpler if we added `bci-base` to `BASE_FIPS_CONTAINERS` if the target is `"dso"`?

dirkmueller:
> Would it maybe make the handling a bit simpler if we added `bci-base` to `BASE_FIPS_CONTAINERS` if the target is `"dso"`?

not really, because we expect `BASE_CONTAINER` to be always existing and exactly one container everywhere. I've pushed an additional commit to do what you're asking for. I think the previous solution was simpler.

dcermak:
> > Would it maybe make the handling a bit simpler if we added `bci-base` to `BASE_FIPS_CONTAINERS` if the target is `"dso"`?
> 
> not really, because we expect `BASE_CONTAINER` to be always existing and exactly one container everywhere. I've pushed an additional commit to do what you're asking for. I think the previous solution was simpler.

You're right, the new solution is pretty complicated. The previous one was better

dcermak (approved):

PR #562, opened 2024-08-13
Mark the kiwi image build test as x86 only

dirkmueller:
this requires significantly more work to pass on non-x86

[CI:TOXENVS] kiwi

### DIFF
--- a/tests/test_kiwi.py
+++ b/tests/test_kiwi.py
@@ -7,6 +7,7 @@ from pytest_container import DerivedContainer
 from pytest_container import container_and_marks_from_pytest_param
 from pytest_container.container import ContainerData
 from pytest_container.container import ImageFormat
+from pytest_container.runtime import LOCALHOST
 
 from bci_tester.data import KIWI_CONTAINERS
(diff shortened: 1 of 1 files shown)


tests/test_kiwi.py
```diff
@@ -53,7 +54,8 @@ def test_kiwi_installation(auto_container):
 
 
 @pytest.mark.skipif(
-    PODMAN_SELECTED and os.geteuid(),
+    LOCALHOST.system_info.arch != "x86_64"
+    or (PODMAN_SELECTED and os.geteuid()),
     # https://github.com/containers/podman/issues/17715#issuecomment-1460227771
     reason="PODMAN requires root privileges for kiwi tests",
```
dcermak, line 60: The reason needs adjustment

dirkmueller: that's ambiguous - it's not a restriction of podman but of how the test is written. it is just a temporary skip, eventually I'll find the time to rework the test to be generic, but we need to get the images out.

alexandrevicenzi (approved):

PR #565, opened 2024-08-14
Java 11 sp6

dirkmueller:


dcermak:
This appears to be still broken, any idea when the fix will be released?

dirkmueller:
> This appears to be still broken, any idea when the fix will be released?

well, I fixed the release pipeline earlier today, so we should be good now.

mgrossu (approved):

PR #569, opened 2024-08-14
[postfix] update tests for changes in logging

Priyankasaggu11929:
Follow up changes in tests, post merge of PR https://github.com/SUSE/BCI-dockerfile-generator/pull/1541

CI tests will give proper signal, when ^ PR is merged, and registry.opensuse.org have refreshed image tags.

### DIFF
--- a/tests/test_postfix.py
+++ b/tests/test_postfix.py
@@ -11,18 +11,9 @@ from bci_tester.data import POSTFIX_CONTAINERS
 
 CONTAINER_IMAGES = POSTFIX_CONTAINERS
 
-POSTFIX_WITH_LOGGING_ENABLED = []
 POSTFIX_WITH_MAILHOG = []
 POSTFIX_WITH_VIRTUAL_MBOX_ENABLED = []
 POSTFIX_WITH_LDAP_ENABLED = []
-
-
-CONTAINERFILE_POSTFIX_WITH_LOGGING_ENABLED = """
-RUN postconf maillog_file=/var/log/postfix.log
-RUN postconf maillog_file_permissions=0644
-HEALTHCHECK --interval=5s --timeout=10s --start-period=30s --retries=3 \
-        CMD postfix status
-"""
 
 
 CONTAINERFILE_POSTFIX_WITH_MAILHOG = """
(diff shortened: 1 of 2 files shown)


dcermak:
@Priyankasaggu11929 can you please rebase?

Priyankasaggu11929:
> @Priyankasaggu11929 can you please rebase?

rebased just now.

tests/test_postfix.py
```diff
@@ -202,26 +177,23 @@ def test_postfix_status(auto_container):
     )
 
 
-@pytest.mark.parametrize(
-    "container_per_test", POSTFIX_WITH_LOGGING_ENABLED, indirect=True
-)
-def test_postfix_send_email_delivered(
-    container_per_test: ContainerData,
-) -> None:
+def test_ppostfix_send_email_delivered(auto_container):
```
dcermak, line 180: Duplicate p. 
Also, shouldn't this use the `auto_container_per_test` fixture as the container is mutated during the test?

```suggestion
def test_postfix_send_email_delivered(auto_container):
```

tests/test_postfix.py
```diff
@@ -231,24 +203,21 @@ def test_postfix_send_email_delivered(
         assert output in log
 
 
-@pytest.mark.parametrize(
-    "container_per_test", POSTFIX_WITH_LOGGING_ENABLED, indirect=True
-)
-def test_postfix_send_email_failed(
-    container_per_test: ContainerData,
-) -> None:
+def test_postfix_send_email_failed(auto_container):
```
dcermak, line 206: Same here, shouldn't this be `auto_container_per_test`?

dcermak (commented):
Now every test includes the following lines:
```python
    auto_container.connection.check_output(
        "postconf maillog_file=/var/log/postfix.log && postfix reload"
    )
```

Shouldn't this become part of the container image instead? Or why is this needed?

Priyankasaggu11929:
> Now every test includes the following lines:
> 
> ```python
>     auto_container.connection.check_output(
>         "postconf maillog_file=/var/log/postfix.log && postfix reload"
>     )
> ```
> 
> Shouldn't this become part of the container image instead? Or why is this needed?

It *was* part of the extended Containefile before this PR, but now with the recent changes from https://github.com/SUSE/BCI-dockerfile-generator/pull/1541 -  postfix container entrypoint script sets  `maillog_file=/dev/stdout` and any override in the extended container doesn't work (because entrypoint still remains the final step when container starts)

Here directing logs to a file is needed in order to check/verify the mail status.

tests/test_postfix.py
Priyankasaggu11929: > Also, shouldn't this use the auto_container_per_test fixture as the container is mutated during the test?
I didn't check (before you pointed out) about this other fixture. Updated it now.

Btw - the mutation here change is the postfix config, and creation of `postfix.log` file?

dcermak: > Btw - the mutation here is change in the postfix config, and creation of `postfix.log` file?

Yes and sending an email. All of these actions modify the running container, which might or might not get reused by other test runs if you use the `container` or `auto_container` fixture. That can result in false positives or false negatives or might only show up in CIs (where very few parallel jobs are run, there container re-use in a single process is much more likely).

dcermak:
> It _was_ part of the extended Containefile before this PR, but now with the recent changes from [SUSE/BCI-dockerfile-generator#1541](https://github.com/SUSE/BCI-dockerfile-generator/pull/1541) - postfix container entrypoint script sets `maillog_file=/dev/stdout` and any override in the extended container doesn't work (because entrypoint still remains the final step when container starts).
> 
> So, I'm trying to update config when container is already started & reload postfix.
> 
> Directing logs to a file is needed in order to check/verify the mail status. We don't have journalctl in the container.

Would it help you to have direct access to the container logs via pytest_container?

Priyankasaggu11929:
> Would it help you to have direct access to the container logs via pytest_container?
> 
> EDIT: I went ahead and implemented it, it turned out to be pretty simple: [dcermak/pytest_container#219](https://github.com/dcermak/pytest_container/pull/219)

Thank you, Dan. Yea, I think direct container logs will help here. I'll test on Monday with the new function, and update.

dcermak (approved):

PR #571, opened 2024-08-16
Limit bci-repo tests to bci-base container

dirkmueller:
When f72a21307efe1b663c88f6f6872db83294a778bc was moved into test_repository it expanded the scope to all known zypper-root based containers. Not all of them are available in TARGET=ibs-cr which broke the repository tests. limit it back to base/base-fips container to unblock the release

[CI:TOXENVS] repository


mgrossu (approved):

dcermak (commented):
Ok as a temporary fix, but this is actually performing valuable tests that are container specific and not repo specific (maybe I should've moved them to all).

PR #572, opened 2024-08-16
Revert "Limit bci-repo tests to bci-base container"

dcermak:
Reverts SUSE/BCI-tests#571

Reminder to re-enable testing on all containers


tests/test_repository.py
```diff
@@ -186,14 +186,8 @@ def test_sle15_packages(container_per_test, pkg):
     )
 
 
-@pytest.mark.skipif(
```
dirkmueller, line 189: this part imho should not be reverted (unless the whole function is moved to all)

dcermak: Yes, done that now

dirkmueller (approved):

Issue #545, opened 2024-07-25
Add tests for 15.7 so that size monitoring for base containers is in place early


dirkmueller:
In 15.6 development phase we learned about major size regressions of the base containers far too late to be able to get them fixed in time. we should enable testing early to have the size monitoring in place.

dirkmueller:
related https://github.com/SUSE/BCI-tests/pull/582 but it needs rebasing (\cc @dcermak )

PR #582, opened 2024-09-02
Enable SP7 testing

dcermak:
This will not work properly before we have the SLE 15 SP7 BCI repo

This fixes https://github.com/SUSE/BCI-tests/issues/545

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -155,8 +155,14 @@ jobs:
             os_version: 15.4
           - toxenv: metadata
             os_version: 15.4
+          - toxenv: all
+            os_version: "15.7"
           - toxenv: base
-            os_version: 15.7
+            os_version: "15.7"
+          - toxenv: minimal
+            os_version: "15.7"
+          - toxenv: metadata
+            os_version: "15.7"
           - toxenv: all
             os_version: "16.0"
           - toxenv: base
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -66,7 +66,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        python_version: ["3.6", "3.9", "3.10", "3.11"]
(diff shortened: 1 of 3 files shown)


dirkmueller:
@dcermak rebase? we need to get this in soonish when 15.7 development starts (I'd be fine with testing all of base/minimal for now, similar to how 16.0 is enabled)

tests/test_all.py
```diff
@@ -376,6 +376,13 @@ def test_no_orphaned_packages(container_per_test: ContainerData) -> None:
 @pytest.mark.parametrize(
     "container", CONTAINERS_WITH_ZYPPER_AS_ROOT, indirect=True
 )
+@pytest.mark.xfail(
```
alexandrevicenzi, line 379: This seems to break tests on 15.4.

.github/workflows/ci.yaml
```diff
@@ -167,6 +165,14 @@ jobs:
             os_version: "16.0"
           - toxenv: kernel_module
             os_version: "16.0"
+          - toxenv: all
+            os_version: "15.7"
```
dirkmueller, line 169: although don't have a strong opinion, I think sorting these before the "16.0" section would be slightly more logical.

tests/test_all.py
dirkmueller: yeah, because they're run on a non-suse host so the container-suseconnect integration is not working. rather than adding a quirk in the tests for this, maybe we should simply not define a repo like in the 16.0 case?

bci_tester/data.py
```diff
@@ -519,7 +519,7 @@ def create_BCI(
 )
 PYTHON311_CONTAINER = create_BCI(
     build_tag="bci/python:3.11",
-    available_versions=_DEFAULT_NONBASE_OS_VERSIONS,
+    available_versions=["15.6", "tumbleweed"],
```
dirkmueller, line 522: python 311 is still available with SP7, why this change?

dirkmueller (approved):

PR #585, opened 2024-09-04
Remove xfail that is no longer relevant

dirkmueller:


dcermak (changes requested):
Breaks in github actions

PR #587, opened 2024-09-06
Reject packages with a SUSE-Firmware license in the BCI repo

dirkmueller:

### DIFF
--- a/tests/test_repository.py
+++ b/tests/test_repository.py
@@ -158,6 +158,26 @@ def test_package_installation(container_per_test, pkg):
 
 
 @pytest.mark.skipif(
+    OS_VERSION == "tumbleweed", reason="No testing for openSUSE"
+)
+@pytest.mark.skipif(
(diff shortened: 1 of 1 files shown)


alexandrevicenzi (dismissed):

dcermak (commented):
LGTM, and can be merged once the CI is green.

tiny nit: atm it's not possible to find out which package is the offender from the logs. This could be worked around by not using `grep -c` and instead only `grep` and then counting the lines in python (should be reasonably cheap)

dcermak:

>> tiny nit: atm it's not possible to find out which package is the offender from the logs. This could be worked around by not using `grep -c` and instead only `grep` and then counting the lines in python (should be reasonably cheap)
>
> doesn't work because the package name is not on the line that is being greped, but somewhere around it. it's not difficult to find the offender, so this should be fine. 

Ah, right! Fair enough then

alexandrevicenzi (approved):

PR #588, opened 2024-09-10
Provide the IBS certificate for checking out the disturl

dcermak:
This test would fail for hosts that don't have internal CA installed, so let's just ship it

[CI:TOXENVS] metadata


dirkmueller (changes requested):
what are you trying to fix? there is no public access on IBS so this doesn't make any sense. 

and there is already a skip of the test for IBS..
https://github.com/SUSE/BCI-tests/pull/588/files#diff-f49ce5cf2ce23f6922b7add17ca149482d190f1c2b5892c5d1d55a28497c5556R468-R470

dcermak:
> what are you trying to fix? there is no public access on IBS so this doesn't make any sense.

I saw a failure in openQA where one of the hosts tried to read the data from IBS, but got a certificate error (and yes, I of course forgot to save the link)

dcermak:
> what are you trying to fix? there is no public access on IBS so this doesn't make any sense.

Found a problematic test run in openQA: https://openqa.suse.de/tests/15452302#step/_root_BCI-tests_metadata_/6

alexandrevicenzi (approved):

PR #591, opened 2024-09-11
Obtain the process cmdline from /proc/ not via ps

dcermak:
This works around the container not shipping ps

[CI:TOXENVS] postfix

CC @Priyankasaggu11929

### DIFF
--- a/tests/test_postfix.py
+++ b/tests/test_postfix.py
@@ -160,16 +160,18 @@ for postfix_ctr in POSTFIX_WITH_VIRTUAL_MBOX_ENABLED:
     )
 
 
-def test_postfix_status(auto_container):
+def test_postfix_status(auto_container: ContainerData):
     """check if Postfix service is running inside the container"""
 
-    # verify PID 1 process is ENTRYPOINT - `/bin/bash /entrypoint/entrypoint.sh postfix start`
-    assert (
-        "/usr/lib/postfix/bin//master -i"
-        in auto_container.connection.check_output(
-            "ps -eo pid,cmd | grep '^ *1 ' | sed 's/^ *1 //'"
-        )
-    )
+    # verify PID 1 process is ENTRYPOINT:
+    # /bin/bash /entrypoint/entrypoint.sh postfix start
+
+    # we don't have ps in the container, so read /proc/1/cmdline instead, it
+    # contains the binary and its arguments separated by \x00
+    cmdline: str = auto_container.connection.file(
+        "/proc/1/cmdline"
+    ).content_string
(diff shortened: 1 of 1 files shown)


Priyankasaggu11929:
LGTM.

(is it related to the SP5 containers (where ps command was dropped some time back?)

dcermak:
Priyanka Saggu ***@***.***> writes:

> LGTM.
>
> (is it related to the SP5 containers (where ps command was dropped some time back?)

Yes, the pipeline started failing tonight as the container image no
longer contains procps in 15SP6

Priyankasaggu11929:
> Priyanka Saggu ***@***.***> writes:
> LGTM. (is it related to the SP5 containers (where ps command was dropped some time back?)
> Yes, the pipeline started failing tonight as the container image no longer contains procps in 15SP6

Thanks Dan. One more follow up question – where was it flagged? (openQA pipelines?)

tests/test_postfix.py
```diff
@@ -160,16 +160,17 @@
...
-            "ps -eo pid,cmd | grep '^ *1 ' | sed 's/^ *1 //'"
-        )
-    )
+    # verify PID 1 process is ENTRYPOINT:
+    # /bin/bash /entrypoint/entrypoint.sh postfix start
+
+    # we don't have ps in the container, so read /proc/1/cmdline instead, it
+    # contains the binary and its arguments split by \x00
+    cmdline: bytes = auto_container.connection.file("/proc/1/cmdline").content
+    proc_1_cmd = cmdline.replace(bytes((0,)), " ".encode()).decode().strip()
```
dirkmueller, line 172: this looks brutally complicated... :/

dcermak:

> @dirkmueller commented on this pull request.
>
>
>
>>      """check if Postfix service is running inside the container"""
>  
> -    # verify PID 1 process is ENTRYPOINT - `/bin/bash /entrypoint/entrypoint.sh postfix start`
> -    assert (
> -        "/usr/lib/postfix/bin//master -i"
> -        in auto_container.connection.check_output(
> -            "ps -eo pid,cmd | grep '^ *1 ' | sed 's/^ *1 //'"
> -        )
> -    )
> +    # verify PID 1 process is ENTRYPOINT:
> +    # /bin/bash /entrypoint/entrypoint.sh postfix start
> +
> +    # we don't have ps in the container, so read /proc/1/cmdline instead, it
> +    # contains the binary and its arguments split by \x00
> +    cmdline: bytes = auto_container.connection.file("/proc/1/cmdline").content
> +    proc_1_cmd = cmdline.replace(bytes((0,)), " ".encode()).decode().strip()
>
> this looks brutally complicated... :/

I've simplified the code as much as possible

dirkmueller (approved):

Issue #504, opened 2024-06-14
Test mariadb version upgrade 


dirkmueller:
Sle 15.5 to 15.6 is going to change the mariadb version and the docker entry point is handling version updates by going through a data migration workflow.

We do not have a test for that and we know the current state was broken because the upstream developers notified us about that issue.

We need to test an upgrade scenario

alexandrevicenzi:
Is the test scenario in mind around [mariadb-upgrade](https://mariadb.com/kb/en/mariadb-upgrade/)?

In theory, SLE is not going to move to a version where `mariadb-upgrade` would not work. If you follow the SLE upgrade paths, you can migrate from SLE 12 MariaDB to SLE 15 with no issues, as long as you always run `mariadb-upgrade` in each SLE migration.

dirkmueller:
@alexandrevicenzi the entrypoint script performs the upgrade itself based on some logic and extra code paths that were broken. we're not testing them. we should be testing them prior to playing out the 15.6 upgrade. 

see `docker_mariadb_upgrade` in the entrypoint.sh

grooverdan:
As your new entrypoint is fixed, I don't think you are in a bad position. The `mariadb-upgrade` checks the file `$DATADIR/mariadb-upgrade-info` ( or mysql_upgrade_info) for the version that was previously installed/upgraded. The old invalid version number in the entrypoint was only used as a pre-check comparison if an upgrade was needed and wasn't actually stored anywhere.

PR #594, opened 2024-09-13
Mariadb upgrade scenario testing

rcmadhankumar:
[CI:TOXENVS] mariadb
fixes: #504

### DIFF
--- a/tests/test_mariadb.py
+++ b/tests/test_mariadb.py
@@ -1,14 +1,21 @@
 """Tests for the MariaDB related application container images."""
 
+import os
 from itertools import product
+from pathlib import Path
+from typing import Any
 from typing import List
 from typing import Optional
+from typing import Union
 
 import pymysql
 import pytest
 from _pytest.mark import ParameterSet
 from pymysql.err import OperationalError
+from pytest_container.container import BindMount
 from pytest_container.container import ContainerData
+from pytest_container.container import ContainerLauncher
+from pytest_container.container import ContainerVolume
 from pytest_container.container import DerivedContainer
 from pytest_container.container import container_and_marks_from_pytest_param
 from pytest_container.pod import Pod
@@ -22,6 +29,8 @@ from tenacity import wait_exponential
 from bci_tester.data import MARIADB_CLIENT_CONTAINERS
 from bci_tester.data import MARIADB_CONTAINERS
 from bci_tester.data import MARIADB_ROOT_PASSWORD
+from bci_tester.data import OS_VERSION
+from bci_tester.runtime_choice import PODMAN_SELECTED
 
 CONTAINER_IMAGES = MARIADB_CONTAINERS
 
@@ -279,3 +288,98 @@ def test_mariadb_healthcheck_galera_cluster_disabled(auto_container_per_test):
     _wait_for_server(conn)
 
     conn.run_expect([1], "healthcheck.sh --su-mysql --galera_online")
+
+
+_DB_ENV = {
+    "MARIADB_USER": _OTHER_DB_USER,
+    "MARIADB_PASSWORD": _OTHER_DB_PW,
+    "MARIADB_DATABASE": _TEST_DB,
+    "MARIADB_ROOT_PASSWORD": MARIADB_ROOT_PASSWORD,
+    "MARIADB_AUTO_UPGRADE": "1",
+}
+
+
+@pytest.mark.parametrize("ctr_image", MARIADB_CONTAINERS)
+@pytest.mark.skipif(
+    OS_VERSION not in ("15.6",),
+    reason="MariaDB upgrade scenario not supported",
+)
+def test_mariadb_upgrade(
+    container_runtime: OciRuntimeBase,
+    pytestconfig: pytest.Config,
+    ctr_image: DerivedContainer,
+    tmp_path: Path,
+    host,
+) -> None:
+    mounts: List[Union[BindMount, ContainerVolume]] = [
+        BindMount(host_path=tmp_path, container_path="/var/lib/mysql")
+    ]
+    mariadb_old = DerivedContainer(
+        base="registry.suse.com/suse/mariadb:10.6",
+        containerfile='RUN set -euo pipefail; head -n -1 /usr/local/bin/gosu > /tmp/gosu;  echo \'exec setpriv --pdeathsig=keep --reuid="$u" --regid="$u" --clear-groups -- "$@"\' >> /tmp/gosu;  mv /tmp/gosu /usr/local/bin/gosu; chmod +x /usr/local/bin/gosu',
+        volume_mounts=mounts,
+        extra_environment_variables=_DB_ENV,
+    )
+    mariadb_new = DerivedContainer(
+        base=ctr_image,
+        volume_mounts=mounts,
+        extra_environment_variables=_DB_ENV,
+    )
+
+    mariadb_cmd = f"mariadb --user={_OTHER_DB_USER} --password={_OTHER_DB_PW} --host=0.0.0.0 {_TEST_DB}"
+
+    def _verify_rowcount(con: Any, table_name: str, no_of_rows: int):
+        rows = (
+            con.check_output(
+                f'echo "SELECT count(*) FROM {table_name};" | {mariadb_cmd}'
+            )
+            .strip()
+            .splitlines()
+        )
+        assert no_of_rows == int(rows[1])
+
+    try:
+        with ContainerLauncher.from_pytestconfig(
+            mariadb_old, container_runtime, pytestconfig
+        ) as launcher:
+            launcher.launch_container()
+            con = launcher.container_data.connection
+            _wait_for_server(con)
+            con.check_output(
+                f'echo "CREATE TABLE random_strings (string VARCHAR(255) NOT NULL);" | {mariadb_cmd}'
+            )
+
+            for _ in range(10):
+                con.check_output(
+                    f"echo 'INSERT INTO random_strings (string) VALUES (MD5(RAND()));' | {mariadb_cmd}"
+                )
+            _verify_rowcount(con, "random_strings", 10)
+
+            con.check_output(
+                f'echo "CREATE TABLE random_numbers (number INT NOT NULL);" | {mariadb_cmd}'
+            )
+            for _ in range(12):
+                con.check_output(
+                    f"echo 'INSERT INTO random_numbers (number) VALUES (RAND());' | {mariadb_cmd}"
+                )
+            _verify_rowcount(con, "random_numbers", 12)
+
+        with ContainerLauncher.from_pytestconfig(
+            mariadb_new, container_runtime, pytestconfig
+        ) as launcher:
+            launcher.launch_container()
+            con = launcher.container_data.connection
+            _wait_for_server(con)
+            _verify_rowcount(con, "random_strings", 10)
+            _verify_rowcount(con, "random_numbers", 12)
+
+    finally:
+        # The tmp_path folder is chown'd to the mariadb user by the mariadb
+        # container, podman remaps that to some subuid that our current user has
+        # no permission to delete. With podman unshare we enter the user
+        # namespace, where we can fix the file permissions so that the cleanup
+        # by pytest works.
+        # Note that we have to chown to root as root inside the user namespace
+        # is our current user
+        if PODMAN_SELECTED and os.getuid() != 0:
+            host.check_output(f"podman unshare chown -R root:root {tmp_path}")


tests/test_mariadb.py
```diff
@@ -265,7 +270,7 @@ def test_mariadb_healthcheck_innodb_initialized(auto_container_per_test):
 
 
 def test_mariadb_healthcheck_galera_cluster_disabled(auto_container_per_test):
-    """
+    """registry.opensuse.org/devel/bci/sle-15-sp6
```
dcermak, line 273:
```suggestion
    """
```

tests/test_mariadb.py
```diff
@@ -276,3 +281,80 @@ def test_mariadb_healthcheck_galera_cluster_disabled(auto_container_per_test):
...
+
+_DB_ENV = {
+    "MARIADB_USER": _OTHER_DB_USER,
+    "MARIADB_PASSWORD": _OTHER_DB_PW,
+    "MARIADB_DATABASE": _TEST_DB,
+    "MARIADB_ROOT_PASSWORD": MARIADB_ROOT_PASSWORD,
+    "MARIADB_AUTO_UPGRADE": "1",
+}
+
+OS_VERSION = os.getenv("OS_VERSION", "")
```
dcermak, line 294:
```suggestion
from bci_tester.data import OS_VERSION
```

tests/test_mariadb.py
```diff
@@ -276,3 +281,80 @@ def test_mariadb_healthcheck_galera_cluster_disabled(auto_container_per_test):
...
+)
+def test_mariadb_upgrade(
+    container_runtime: OciRuntimeBase,
+    pytestconfig: pytest.Config,
+    ctr_image: DerivedContainer,
+    tmp_path: Path,
+) -> None:
+    mounts = [BindMount(host_path=tmp_path, container_path="/var/lib/mysql")]
+    mariadb_old = Container(
+        url="registry.opensuse.org/devel/bci/sle-15-sp6/containerfile/suse/mariadb:latest",
```
dcermak, line 309: please use the mariadb container from `registry.suse.com/suse/mariadb`, at a version that is different from the containers in `MARIADB_CONTAINER`.

tests/test_mariadb.py
```diff
@@ -276,3 +281,80 @@ def test_mariadb_healthcheck_galera_cluster_disabled(auto_container_per_test):
...
+
+    with ContainerLauncher.from_pytestconfig(
+        mariadb_new, container_runtime, pytestconfig
+    ) as launcher:
+        launcher.launch_container()
+        con = launcher.container_data.connection
+        _wait_for_server(con)
+        expected_version = mariadb_new.url.split(":")[1]
+        if expected_version != "latest":
+            assert expected_version in con.check_output("mysql --version")
```
dcermak, line 350: I don't get the point of this check to be honest

rcmadhankumar: this is just an additional check that the upgraded is done successfully. But, yes it is not mandatory.
It is expected that the mariadb image:x.y to run Mariadb-Vx.y.

rcmadhankumar: we can ignore this check.

dcermak: I don't understand what a version check has to do with the upgrade?

rcmadhankumar: this is just a check that our database is upgraded to newer version.

tests/test_mariadb.py
rcmadhankumar: resolved

tests/test_mariadb.py
rcmadhankumar: fixed.

tests/test_mariadb.py
rcmadhankumar: version 10.6 from `registry.suse.com/suse/mariadb` is added.

tests/test_mariadb.py
rcmadhankumar: removed this section as it doesn't make much sense.

tests/test_mariadb.py
```diff
@@ -276,3 +280,76 @@ def test_mariadb_healthcheck_galera_cluster_disabled(auto_container_per_test):
...
+)
+def test_mariadb_upgrade(
+    container_runtime: OciRuntimeBase,
+    pytestconfig: pytest.Config,
+    ctr_image: DerivedContainer,
+    tmp_path: Path,
+) -> None:
+    mounts = [BindMount(host_path=tmp_path, container_path="/var/lib/mysql")]
+    mariadb_old = DerivedContainer(
+        base="registry.suse.com/suse/mariadb:10.6",
```
alexandrevicenzi, line 306: There's no guarantee that you will be able to migrate from 10.6 to newer MariaDB containers.

I believe the test should include a mapping from which version to which newer version, otherwise this test could fail on a new container image, not because it is broken, but because MariaDB does not support migrating 10.6 to version X.Y.

So instead of `MARIADB_CONTAINERS` it should be passed something like `MARIADB_CONTAINERS_UPGRADE_VERSIONS`, which would be a tuple, or another map style type.

tests/test_mariadb.py
```diff
@@ -276,3 +280,76 @@ def test_mariadb_healthcheck_galera_cluster_disabled(auto_container_per_test):
...
+        launcher.launch_container()
+        con = launcher.container_data.connection
+        _wait_for_server(con)
+        mariadb_cmd = f"mariadb --user={_OTHER_DB_USER} --password={_OTHER_DB_PW} --host=0.0.0.0 {_TEST_DB}"
+
+        con.check_output(
+            f'echo "CREATE TABLE test (id serial PRIMARY KEY, num integer, data varchar(32));" | {mariadb_cmd}'
+        )
+        con.check_output(
+            f"echo 'INSERT INTO test (num, data) VALUES (100, \"abcdef\")' | {mariadb_cmd}"
```
alexandrevicenzi, line 328: According to docs:

- Updating the system tables in the mysql database to the newest version. This is very quick.
- [mariadb-upgrade](https://mariadb.com/kb/en/mariadb-upgrade/) also runs [mariadb-check --check-upgrade](https://mariadb.com/kb/en/mariadb-check/) to check if there have been any collation changes between the major versions. This recreates indexes in old tables that are using any of the changed collations. This can take a bit of time if there are a lot of tables or there are many tables which used the changed collation. The last time a collation changed was in MariaDB/MySQL 5.1.23.

It would be nice to touch system tables, such as changing the database charset to a non-standard one.

It would be nice to have a data file that has multiple commands and tables. Ideally creating indexes, so we know that MariaDB can migrate indexes, plus perhaps including more data types than just integer and varchar.

Regular tables are not very likely to break or change unless there's a change in the data type of a column I believe.

rcmadhankumar: Do we have any existing SQL file that has a good load of data(we might have used in earlier experiments) or would you like to recommend any specific dataset for this or we can create our own random data(with indexes for sure)?

alexandrevicenzi: I don't have any, maybe the maintainer has one. Otherwise, we could grab some dataset from Kaggle if needed.

There's no need to be a big dump, just having a few tables, indexes, and changing things that touch system tables would be sufficient I believe.

tests/test_mariadb.py
rcmadhankumar: @alexandrevicenzi as per the docs here: https://mariadb.com/kb/en/upgrading-between-major-mariadb-versions/ we should be able to upgrade between any 2 major versions of mariadb.

tests/test_mariadb.py
rcmadhankumar: Additional info: 

As per the official mariadb upgrade tests here: https://github.com/MariaDB/server/blob/fd0cc2b1fdee24955d39cdcecbc457324b85b2d2/tests/upgrade_from/test_upgrade.sh, they create a [simple dataset ](https://github.com/MariaDB/server/blob/fd0cc2b1fdee24955d39cdcecbc457324b85b2d2/tests/upgrade_from/test_upgrade.sh#L213)for testing the upgrade.

So, i think we don't have to create dataset specific to each versions and the dataset can be common across all the version. But, as you suggested the dataset can have indices and we can also touch the system tables.

tests/test_mariadb.py
```diff
@@ -276,3 +280,108 @@ def test_mariadb_healthcheck_galera_cluster_disabled(auto_container_per_test):
...
+
+        rows = (
+            con.check_output(f'echo "SELECT count(*) FROM random_strings;" | {mariadb_cmd}')
+            .strip()
+            .splitlines()
+        )
+
+        assert rows and len(rows) == 2
+        _, count = rows[-1].split()
+        assert count == "10"
```
dcermak, line 341: `SELECT count(*) FROM random_strings;` returns the following:
```
count(*)
10
```

i.e. the `strip().splitlines()` yields: ["count(*)", "10"], hence the assert on `rows and len(rows) == 2` passes, but to obtain count just run: `count = rows[-1]`.

This also needs to be changed in the four places below (you might want to put this into a separate function)

tests/test_mariadb.py
```diff
@@ -276,3 +280,108 @@ def test_mariadb_healthcheck_galera_cluster_disabled(auto_container_per_test):
...
+@pytest.mark.skipif(
+    OS_VERSION in ("15.5",), reason="MariaDB upgrade scenario not supported"
+)
+def test_mariadb_upgrade(
+    container_runtime: OciRuntimeBase,
+    pytestconfig: pytest.Config,
+    ctr_image: DerivedContainer,
+    tmp_path: Path,
+) -> None:
+    mounts = [BindMount(host_path=tmp_path, container_path="/var/lib/mysql")]
```
dcermak, line 304: nit to make the type checker happy:
```suggestion
    mounts: List[Union[BindMount, ContainerVolume]] = [BindMount(host_path=str(tmp_path), container_path="/var/lib/mysql")]
```

tests/test_mariadb.py
```diff
@@ -276,3 +280,108 @@ def test_mariadb_healthcheck_galera_cluster_disabled(auto_container_per_test):
...
+
+        rows = (
+            con.check_output(f'echo "SELECT count(*) FROM random_numbers;" | {mariadb_cmd}')
+            .strip()
+            .splitlines()
+        )
+
+        assert rows and len(rows) == 2
+        _, count = rows[-1].split()
+        assert count == "12"
```
dcermak, line 387: You can append the following cleanup code to fix the file permissions on `tmp_path` for rootless podman:
```Python
        # The tmp_path folder is chown'd to the mariadb user by the mariadb
        # container, podman remaps that to some subuid that our current user has
        # no permission to delete. With podman unshare we enter the user
        # namespace, where we can fix the file permissions so that the cleanup
        # by pytest works.
        # Note that we have to chown to root as root inside the user namespace
        # is our current user
        if PODMAN_SELECTED and geteuid() != 0:
            host.check_output(f"podman unshare chown -R root {tmp_path}")
```

I'd suggest to wrap the whole thing in a `try … except` block though, otherwise this will not get executed on failure. Ideally we'd move this to a separate fixture though

tests/test_mariadb.py
```diff
@@ -276,3 +285,101 @@ def test_mariadb_healthcheck_galera_cluster_disabled(auto_container_per_test):
...
+        mariadb_new, container_runtime, pytestconfig
+    ) as launcher:
+        launcher.launch_container()
+        con = launcher.container_data.connection
+        _wait_for_server(con)
+        _verify_rowcount(con, "random_strings", 10)
+        _verify_rowcount(con, "random_numbers", 12)
+
+        # to change tmp_path permissions
+        host.check_output(f"podman unshare chown -R root:root {tmp_path}")
```
dcermak, line 385:
```suggestion
    host.check_output(f"podman unshare chown -R root:root {tmp_path}")
```

tests/test_mariadb.py
rcmadhankumar: fixed.

tests/test_mariadb.py
rcmadhankumar: fixed.

tests/test_mariadb.py
rcmadhankumar: added this code in try and finally blocks.

tests/test_mariadb.py
rcmadhankumar: Right now, the test case written only to verify the upgrade scenario for SP6, as the URL for old mariadb container is hard coded in the test case we are yet to customise the test case to choose the url for old container based on OS_VERSION.
Will update the test case.

dcermak (dismissed):

tests/test_mariadb.py
rcmadhankumar: We have decided to use the standard data set used in official mariadb test cases.

alexandrevicenzi (approved):

tests/test_mariadb.py
rcmadhankumar: right now the test case will run for only 15.6.

dcermak (approved):

alexandrevicenzi (approved):

PR #605, opened 2024-10-02
Skip manifest test on ancient podman versions

dirkmueller:
[CI:TOXENVS] metadata

### DIFF
--- a/tests/test_metadata.py
+++ b/tests/test_metadata.py
@@ -87,6 +87,7 @@ from bci_tester.data import RUST_CONTAINERS
 from bci_tester.data import SPACK_CONTAINERS
 from bci_tester.data import TOMCAT_CONTAINERS
 from bci_tester.data import ImageType
+from bci_tester.runtime_choice import PODMAN_SELECTED
 
 #: The official vendor name
 VENDOR = "openSUSE Project" if OS_VERSION == "tumbleweed" else "SUSE LLC"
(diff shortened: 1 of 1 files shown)


dcermak (dismissed):

tests/test_metadata.py
```diff
@@ -647,6 +648,9 @@ def test_reference(
 
 
 @SKIP_IF_TW_MARK
+@pytest.mark.skipif(
+    not PODMAN_SELECTED, reason="we only need to test it for one runtime"
```
grisu48, line 652:
```suggestion
    not PODMAN_SELECTED, reason="we only need to test it for podman"
```
Nitpicking: I would make the error message more explicit about the underlying reason for better readability in the error case.

grisu48 (dismissed):

grisu48:
A related failure that would be fixed is e.g. https://openqa.suse.de/tests/15579984#step/_root_BCI-tests_metadata_/2

On 15-SP2 the podman version is 2.1.1, on 15-SP3 it is 3.4.7

tests/test_metadata.py
dirkmueller: well, there is nothing specific to podman in the test, so that improved reasoning makes it imho actually worse. anyhow, I think I addressed your concern in the followup.

grisu48:
We have already some test runs that are passing, see e.g.

image

Do we need to wait for the full CI to complete or can we proceed with this one?

alexandrevicenzi (approved):

grisu48 (approved):

PR #606, opened 2024-10-02
Remove use of git in tests

dirkmueller:
This is doing deep clones which takes forever. also the zypper installs prevent us from going with leaner base containers.


dcermak (approved):

grisu48 (approved):
LGTM!

PR #607, opened 2024-10-05
Remove usage of zypper

dirkmueller:
There is no zypper anymore in bci micro based containers.

### DIFF
--- a/tests/test_postfix.py
+++ b/tests/test_postfix.py
@@ -28,13 +28,9 @@ HEALTHCHECK --interval=5s --timeout=10s --start-period=30s --retries=3 \
 
 
 CONTAINERFILE_POSTFIX_WITH_LDAP_ENABLED = """
-RUN set -euo pipefail; \
-zypper -n in --no-recommends find openldap2 openldap2-client; \
-zypper -n clean; \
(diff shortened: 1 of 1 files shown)


dcermak (changes requested):
This fails on Tumbleweed with:
```
0.110 /bin/sh: line 1: tar: command not found
```

You could try to use the `container_git_clone` fixture instead: https://github.com/SUSE/BCI-tests/blob/35aefbb4f8c646cc9818439601412dec4f295f9a/conftest.py#L22

dirkmueller:
> You could try to use the `container_git_clone` fixture instead:
> 
> https://github.com/SUSE/BCI-tests/blob/35aefbb4f8c646cc9818439601412dec4f295f9a/conftest.py#L22

that does the cloning at test runtime, rather than at build time of the container. if we don't want to add tar to the container on tumbleweed then I think I need to wait until the multibuild refactoring is over and make this a proper multibuild package.

dirkmueller:
see https://github.com/SUSE/BCI-dockerfile-generator/pull/1817 instead.

dirkmueller:
thanks for merging the tar change. now I'd like to get this change reviewed as otherwise the tests are broken.

dcermak (approved):

PR #608, opened 2024-10-07
Correct the skipif in test_kiwi_create_image to exclude docker

dcermak:
[CI:TOXENVS] kiwi


alexandrevicenzi (dismissed):

tests/test_kiwi.py
```diff
@@ -55,7 +55,10 @@ def test_kiwi_installation(auto_container):
     reason="test is atm x86_64 specific",
 )
 @pytest.mark.skipif(
-    PODMAN_SELECTED and os.geteuid() != 0,
+    PODMAN_SELECTED, reason="Kiwi containers are only supported with podman"
```
dirkmueller, line 58: shouldn't that be `DOCKER_SELECTED` then? otherwise the reason does not match the condition.

dirkmueller: or `not PODMAN_SELECTED`

dcermak: My brain is not braining anymore :melting_face:

dirkmueller (approved):

tests/test_kiwi.py
```diff
@@ -55,7 +55,10 @@ def test_kiwi_installation(auto_container):
     reason="test is atm x86_64 specific",
 )
 @pytest.mark.skipif(
-    PODMAN_SELECTED and os.geteuid() != 0,
+    DOCKER_SELECTED, reason="Kiwi containers are only supported with podman"
```
dirkmueller, line 58: what's the reason for excluding docker? any rationale? there is nothing in the commit message, no reference here in the reason either. I was searching my emails and slack, and couldn't find anything. why are we doing this?

dirkmueller: the previous condition was trying to avoid rootless podman. both docker and rootful podman are supposed to work from what I can say..

dirkmueller: maybe you want https://github.com/SUSE/BCI-tests/pull/610 instead ?

PR #621, opened 2024-10-22
Drop psycopg2 dependency

dirkmueller:
This is obsolete (replacement is psycopg) and is a non-wheel binary extension which means the host needs to have postgresql-server-devel and python-devel installed, which is painful for tesing on minified distributions. Use a pure-python implementation instead.

### DIFF
--- a/tests/test_postgres.py
+++ b/tests/test_postgres.py
@@ -4,7 +4,7 @@ from itertools import product
 from typing import List
 from typing import Optional
 
-import psycopg2
+import pg8000.dbapi
 import pytest
 from _pytest.mark import ParameterSet
 from pytest_container.container import ContainerData
@@ -94,11 +94,7 @@ def test_postgres_db_env_vars(
     variables.
 
     """
-    conn = None
-    cur = None
-
     pgdata = container_per_test.inspect.config.env["PGDATA"]
-
     pgdata_f = container_per_test.connection.file(pgdata)
     assert pgdata_f.exists
     assert pgdata_f.user == "postgres"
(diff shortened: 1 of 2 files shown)


dcermak:
We definitely must test this with Python 3.6 as pg8000 hasn't had a release supporting Python 3.6 for over two years

dirkmueller:
I'll test this. Anyhow,, newerr python versions are available, it's just a bit of setup work.

Also we can pin an older version for 3.6.

dirkmueller:
> I'll test this. Anyhow,, newerr python versions are available, it's just a bit of setup work.
> 
> Also we can pin an older version for 3.6.

I've tested it, it works fine on 3.6. just to avoid any concerns, I've pinned an older version for 3.6.

dirkmueller:
actually the pinning breaks it.

grisu48:
@dirkmueller when you think this is ready, let me try to get this running on a 12-SP5 host. Those are AFAICT the most problematic ones with the dependencies.

dirkmueller:
> @dirkmueller when you think this is ready, let me try to get this running on a 12-SP5 host. Those are AFAICT the most problematic ones with the dependencies.

I tested on a SLE15 SP6 with python 3.6, feel free to test on sle12 also, but I think we could maybe convince PM that we don't need to test SLE12 anymore.

grisu48:
@dirkmueller Works on 12-SP5 with `pip3.6 --quiet install 'pg8000<1.27.0'`. I will prepare a PR for updating https://github.com/os-autoinst/os-autoinst-distri-opensuse accordingly as well and link it here. Both need to be merged together.

dirkmueller:
@grisu48 tox should install those for you, you don't need to do that with pip. I agree it would be nice to remove the installation of the additional packages but that can be done at any later point after merging this.

grisu48:
> @grisu48 tox should install those for you, you don't need to do that with pip. I agree it would be nice to remove the installation of the additional packages but that can be done at any later point after merging this.

Yeah, I realized the same. Will still prepare the PR for removing the not needed dependencies but they are not needed for this PR.

grisu48 (approved):

PR #624, opened 2024-10-23
Run unit-tests for py 313

dirkmueller:


dirkmueller:
tested locally, cannot be enabled in CI yet because of missing python:3.13 bci container.

PR #628, opened 2024-10-28
Add openjdk11 to the testing matrix

dirkmueller:


dcermak:
Fails with:
```
'apache-tomcat_10.1-openjdk11' not found in `markers` configuration option
```
-> needs a `pyproject.toml` entry

PR #631, opened 2024-10-29
Enable repository testing for SP7

dirkmueller:


dirkmueller:
there is a small issue with the repo (as seen in the all failure) but that is being fixed.

dirkmueller:
merging as the repo became available..

PR #637, opened 2024-10-31
run all the jobs for sp7

dirkmueller:
probably a bit too early to enable it, so just as a test that can be manually triggered for now.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -124,6 +124,7 @@ jobs:
         toxenv: ${{fromJson(needs.gentestmatrix.outputs.matrix)}}
         os_version:
           - 15.6
+          - 15.7
           - "tumbleweed"
         include:
           - toxenv: fips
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yaml
```diff
@@ -172,20 +173,6 @@ jobs:
             os_version: 15.4
           - toxenv: metadata
             os_version: 15.4
-          - toxenv: all
```
grisu48, line 175: Accidental removal?

grisu48 (changes requested):

.github/workflows/ci.yaml
dcermak: no it's not accidental, the `all` environment should be running by default already

alexandrevicenzi: nit: it would be nice to split into another commit with proper justification.

alexandrevicenzi (dismissed):

grisu48 (dismissed):
LGTM!

.github/workflows/ci.yaml
```diff
@@ -125,6 +125,7 @@ jobs:
         os_version:
           - 15.5
```
grisu48, line 126: Leap 15.5 is EOL since End of December 2024. Let's remove it as well.

dirkmueller: good point, added another commit

dcermak (approved):

grisu48 (approved):

PR #638, opened 2024-10-31
Check whether the nssdb was created in fips mode in the fips containers

dirkmueller:
we are now enforcing fips mode for NSS, so we don't need to modify the mode. just check for the mode being correct.

Also fix password handling hang.


pablo-herranz (approved):

pablo-herranz:
LGTM! :rocket:

PR #644, opened 2024-11-12
Add initial testing for the SUSE AI containers

dirkmueller:

### DIFF
--- a/tests/test_ai.py
+++ b/tests/test_ai.py
@@ -1,0 +1,59 @@
+"""This module contains the tests for the SUSE AI container."""
+
+import pytest
+import requests
+from tenacity import retry
+from tenacity import stop_after_attempt
+from tenacity import wait_exponential
+
+from bci_tester.data import OLLAMA_CONTAINER
+from bci_tester.data import OPENWEBUI_CONTAINER
+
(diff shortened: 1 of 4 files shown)


dcermak (commented):
LGTM, but the way I see it, we're not testing it in github actions at all? So There's really no way for us to not accidentally break their testing pipeline.

tests/test_ai.py
```diff
@@ -0,0 +1,59 @@
...
+    # Retry 5 times with exponential backoff delay
+    @retry(
+        wait=wait_exponential(multiplier=1, min=4, max=10),
+        stop=stop_after_attempt(5),
+    )
+    def check_ollama_response():
+        resp = requests.get(f"http://localhost:{host_port}/", timeout=30)
+        resp.raise_for_status()
+        assert "Ollama is running" in resp.text
+
```
grisu48, line 36: For now OK, I suggest to add another test that performs a query against ollama and not only check if the default page works.

grisu48 (approved):
Improvement suggestion for the future, for now nothing to be done.

LGTM.

alexandrevicenzi (commented):
LGTM

PR #646, opened 2024-11-12
Temporarily skip startup test on ppc64le

grisu48:
Due to worker unavailability we need to run ppc64le on emulated workers at the moment where the startup times are always larger than the threshold. Therefore there is no point in running this test at all on ppc64le at the moment.

[CI:TOXENVS] init

### DIFF
--- a/tests/test_init.py
+++ b/tests/test_init.py
@@ -44,6 +44,12 @@ class TestSystemd:
     def test_systemd_boottime(self, auto_container):
         """Ensure the container startup time is below 5 seconds"""
 
+        # FIXME: Temporary workaround: While using emulated workers, the startup time test doesn't make sense.'
+        if auto_container.connection.system_info.arch == "ppc64le":
+            pytest.skip(
+                "boottime test temporarily disabled on emulated ppc64le workers."
+            )
+
         # Container startup time limit in seconds - aarch64 workers are slow sometimes.
         startup_limit = (
             11
--- a/bci_tester/runtime_choice.py
+++ b/bci_tester/runtime_choice.py
@@ -1,3 +1,5 @@
+"""Provide helper variables to determine which runtime is used for this test run."""
+
 from pytest_container import get_selected_runtime
 
 #: Is :command:`docker` the selected runtime?


tests/test_init.py
```diff
@@ -44,6 +44,10 @@ def test_systemd_no_udev_present(self, auto_container):
     def test_systemd_boottime(self, auto_container):
         """Ensure the container startup time is below 5 seconds"""
 
+        # Temporary workaround: While using emulated workers, the startup time test doesn't make sense.'
+        if auto_container.connection.system_info.arch == "ppc64le":
+            return
```
dcermak, line 49: Please use `pytest.skip()` with a reason why here
Also, please add a giant `FIXME` somewhere

grisu48: Done.

tests/test_init.py
```diff
@@ -44,6 +44,13 @@ def test_systemd_no_udev_present(self, auto_container):
     def test_systemd_boottime(self, auto_container):
         """Ensure the container startup time is below 5 seconds"""
 
+        # FIXME: Temporary workaround: While using emulated workers, the startup time test doesn't make sense.'
+        if auto_container.connection.system_info.arch == "ppc64le":
+            pytest.skip(
+                "boottime test temporarily disabled on emulated ppc64le workers."
+            )
+            return
```
dcermak, line 52: Bad wording on my end, you can then omit the `return`, everything after `pytest.skip()` is ignored.
```suggestion
```

tests/test_init.py
alexandrevicenzi: Can this FIXME have an issue so we dont forget about it?

tests/test_init.py
grisu48: done

tests/test_init.py
grisu48: https://github.com/SUSE/BCI-tests/issues/647

dirkmueller (dismissed):

dirkmueller (approved):

PR #648, opened 2024-11-13
Remove node 20 from SP7 and Tumbleweed

dirkmueller:
It has been replaced by 22 in both.


alexandrevicenzi (approved):
There are some build failures, but they do not look related to these changes.

PR #650, opened 2024-11-13
Temporarily increase timeouts

grisu48:
Increase the timeouts for container startup temporarily due to the need of using emulated ppc workers.

[CI:TOXENVS] postgres,mariadb

### DIFF
--- a/tests/test_postgres.py
+++ b/tests/test_postgres.py
@@ -11,6 +11,7 @@ from _pytest.mark import ParameterSet
 from pytest_container.container import ContainerData
 from pytest_container.container import DerivedContainer
 from pytest_container.container import container_and_marks_from_pytest_param
+from pytest_container.runtime import LOCALHOST
 
 from bci_tester.data import POSTGRESQL_CONTAINERS
 from bci_tester.data import POSTGRES_PASSWORD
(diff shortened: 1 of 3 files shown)


tests/test_postgres.py
```diff
@@ -87,6 +87,7 @@ def test_postgres_db_env_vars(
     pg_user: Optional[str],
     password: str,
     username: Optional[str],
+    healthcheck_timeout: 180,  # FIXME https://github.com/SUSE/BCI-tests/issues/647
```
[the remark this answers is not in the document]
grisu48: Crap, yes indeed.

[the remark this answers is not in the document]
grisu48: Fixed.

tests/test_postgres.py
```diff
@@ -87,6 +88,9 @@ def test_postgres_db_env_vars(
     pg_user: Optional[str],
     password: str,
     username: Optional[str],
+    healthcheck_timeout: timedelta(
+        seconds=180
```
dirkmueller, line 92: that could be `minutes=3` also

tests/test_postgres.py
```diff
@@ -87,6 +88,9 @@ def test_postgres_db_env_vars(
...
+    ),  # FIXME https://github.com/SUSE/BCI-tests/issues/647
```
dirkmueller, line 93:
```suggestion
    ) if LOCALHOST.system_info.arch == "ppc64le" else None ,  # FIXME https://github.com/SUSE/BCI-tests/issues/647
```

grisu48: Please check again @dirkmueller , this was the wrong place for the `healthcheck_timeout` and in the outer scope `LOCALHOST` is not available.

dirkmueller (approved):

PR #653, opened 2024-11-13
Switch cosign container to L3 supported

dirkmueller:
[CI:TOXENVS] metadata


dirkmueller:
needed to unbreak the CI

Issue #647, opened 2024-11-12
Remove temporary boottime workaround


grisu48:
In https://github.com/SUSE/BCI-tests/pull/646 we introduced the workaround to skip the boottime test on ppc64le due to openQA worker problems. This workaround should be removed once we have the workers back.

grisu48:
https://github.com/SUSE/BCI-tests/pull/650

mgrossu:
More PRs to be reverted:

https://github.com/SUSE/BCI-tests/pull/655

grisu48:
FYI - We could move most BCI tests back to actual PowerPC workers, but still need emulated workers for SLES16, so this is still WIP.

PR #655, opened 2024-11-14
Increase timeouts postgres

mgrossu:
A higher increase of the timeouts.

[CI:TOXENVS] postgres

### DIFF
--- a/bci_tester/data.py
+++ b/bci_tester/data.py
@@ -727,6 +727,12 @@ POSTGRESQL_CONTAINERS = [
         available_versions=pg_versions,
         forwarded_ports=[PortForwarding(container_port=5432)],
         extra_environment_variables={"POSTGRES_PASSWORD": POSTGRES_PASSWORD},
+        # https://github.com/SUSE/BCI-tests/issues/647
+        healthcheck_timeout=(
+            timedelta(minutes=6)
+            if LOCALHOST.system_info.arch == "ppc64le"
+            else None
+        ),
     )
     for pg_ver, pg_versions in (
         (14, ["tumbleweed"]),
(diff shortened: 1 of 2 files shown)


tests/test_postgres.py
```diff
@@ -70,7 +70,7 @@ def _generate_test_matrix() -> List[ParameterSet]:
                         ),
                         # https://github.com/SUSE/BCI-tests/issues/647
                         healthcheck_timeout=(
-                            timedelta(minutes=3)
+                            timedelta(minutes=4)
```
grisu48, line 73: If 3 was not enough, then yank it up to 6. If we are on the brick this will only lead to sporadic failures.

grisu48:
Please add a link to this PR to https://github.com/SUSE/BCI-tests/issues/647#issuecomment-2473548314, so that we can track which PRs we need to revert once we have the actual PPC workers back.

bci_tester/data.py
```diff
@@ -727,6 +727,12 @@ def create_BCI(
         available_versions=pg_versions,
         forwarded_ports=[PortForwarding(container_port=5432)],
         extra_environment_variables={"POSTGRES_PASSWORD": POSTGRES_PASSWORD},
+        # FIXME: https://github.com/SUSE/BCI-tests/issues/647
```
dirkmueller, line 730: please remove FIXME, that just lowers the pylint scoring and somebody else needs to compensate for that.

dirkmueller (approved):

tests/test_postgres.py
mgrossu: Increased to 6 minutes.

bci_tester/data.py
mgrossu: Removed it.

PR #657, opened 2024-11-14
Add startup delay of 1min to git healthcheck

grisu48:
Container startup is too fast for the emulated ppc64 workers.

* Related failure: https://openqa.suse.de/tests/15931383#step/bci_test_podman/59

[CI:TOXENVS] git

### DIFF
--- a/tests/test_git.py
+++ b/tests/test_git.py
@@ -6,6 +6,7 @@ from pytest_container.container import ImageFormat
 from pytest_container.container import container_and_marks_from_pytest_param
 from pytest_container.pod import Pod
 from pytest_container.pod import PodData
+from pytest_container.runtime import LOCALHOST
 
 from bci_tester.data import GIT_CONTAINER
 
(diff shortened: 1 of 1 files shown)


tests/test_git.py
```diff
@@ -56,7 +56,7 @@
 
 CMD ["/usr/sbin/sshd", "-De"]
 
-HEALTHCHECK --interval=5s --timeout=5s --retries=5 CMD ["/usr/bin/ssh-keyscan", "-vvv", "-H", "-p", "{_SSH_PORT}", "127.0.0.1"]
+HEALTHCHECK --interval=5s --timeout=5s --retries=5 --start-period=1m CMD ["/usr/bin/ssh-keyscan", "-vvv", "-H", "-p", "{_SSH_PORT}", "127.0.0.1"]
```
dirkmueller, line 59: can you make this ppc64le specific ? I think 1m is unreasonable otherwise.

grisu48: I can, but in this case I thought that we do not care about the start period as a test subject.

grisu48: Amended, but IMHO this looks now really ugly.

dirkmueller (approved):

tests/test_git.py
dirkmueller: we do care about startup time because in a kubernetes environment the container will fail to start if it isn't ready within the startup time configured (which by default is rather low)

PR #661, opened 2024-11-15
Add xfail to rust-openssl test

dcermak:
The tests of the rust-openssl crate are currently failing because upstream enabled tests using the prime192 curve, but this curve has been disabled in openSUSE.

Pending upstream fix: https://github.com/sfackler/rust-openssl/pull/2330

[CI:TOXENVS] rust


dcermak:
No longer required as upstream was very, very quick

PR #667, opened 2024-11-25
Introduce manual TARGET

grisu48:
When the TARGET is set to manual, the tester should take a given
CONTAINER_URL without modifying it.


dcermak (commented):
This is not necessarily a good idea. With this change, you can now no longer test images built with both kiwi and docker in one run

grisu48:
> This is not necessarily a good idea. With this change, you can now no longer test images built with both kiwi and docker in one run

We need this when setting custom image URLs that don't want to have auto-generated suffixes, see e.g. https://openqa.suse.de/tests/16000117#step/_root_BCI-tests_repository_/1

But I'm open to add a different target like `manual` and leave `custom` to how it works as of now.

dirkmueller:
> But I'm open to add a different target like manual and leave custom to how it works as of now.

how about simply releasing sle15-image to ToTest? that would solve it without any code modifications on the qa pipeline.

grisu48:
> > But I'm open to add a different target like manual and leave custom to how it works as of now.
> 
> how about simply releasing sle15-image to ToTest? that would solve it without any code modifications on the qa pipeline.

Yeah that's better here, but the ability to manually configure the container under test still appears like a good feature to have.

dcermak (commented):
I am not entirely sold on this PR tbh. So far we have been using `baseurl` as the URL to the project where all your containers reside and let the test code figure out the correct url + build tag. With this change, you now have to set the url to the container directly, so the term `BASEURL` doesn't make a lot of sense anymore and it changes the existing behavior completely.

Have you considered using a different environment variable for the case `TARGET=manual`?

grisu48:
> Have you considered using a different environment variable for the case `TARGET=manual`?

I now introduced the variable `CONTAINER_URL` to be used with `TARGET=manual`.

dcermak (dismissed):

bci_tester/data.py
```diff
@@ -350,7 +350,11 @@ def create_BCI(
     ):
         marks.append(pytest.mark.__getattr__(build_tag_base.replace(":", "_")))
 
-    if OS_VERSION == "tumbleweed":
+    if TARGET == "manual":
+        baseurl = os.getenv("CONTAINER_URL")
+        if not baseurl:
+            raise ValueError("Missing CONTAINER_URL for TARGET manual")
```
dirkmueller, line 356: this needs to be allowlisted in `tox.ini` . 

how is this different to TARGET=custom with BASEURL=xxx ?

grisu48: With `TARGET=custom` and `BASEURL=foobar` BCI-Test will still apply suffixes on the BASEURL.

dcermak (approved):

PR #672, opened 2024-11-26
Enable AI containers for tox -e all

dcermak:
[CI:TOXENVS] all


dirkmueller (changes requested):
AssertionError: IMAGES_AND_NAMES must have all containers from ALL_CONTAINERS

now we need to update tests/test/metadata as well

dirkmueller (approved):

PR #675, opened 2024-11-27
Add testing for the python 3.9 SAC container

dirkmueller:

### DIFF
--- a/bci_tester/data.py
+++ b/bci_tester/data.py
@@ -260,6 +260,7 @@ class ImageType(enum.Enum):
     """
 
     LANGUAGE_STACK = enum.auto()
+    SAC_LANGUAGE_STACK = enum.auto()
     APPLICATION = enum.auto()
     SAC_APPLICATION = enum.auto()
     OS = enum.auto()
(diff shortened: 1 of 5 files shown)


bci_tester/data.py
```diff
@@ -355,7 +354,7 @@ def create_BCI(
         if not baseurl:
             raise ValueError("Missing CONTAINER_URL for TARGET manual")
     elif OS_VERSION == "tumbleweed":
-        if bci_type in (ImageType.APPLICATION, ImageType.SAC_APPLICATION):
+        if bci_type in (ImageType.APPLICATION, ImageType.SUSE_APPLICATION):
```
dcermak, line 357:
```suggestion
        if bci_type in (ImageType.APPLICATION, ImageType.SUSE_APPCOLLECTION):
```

dcermak (dismissed):
You have to change the python 3.9 container to `ImageType.SUSE_APPCOLLECTION` in [`tests/test_metadata.py:166`](https://github.com/SUSE/BCI-tests/pull/675/files#diff-f49ce5cf2ce23f6922b7add17ca149482d190f1c2b5892c5d1d55a28497c5556R166), otherwise the reference & url tests fail

dcermak (commented):
LGTM, but now we have a new problem: we have two different python 3.11 containers with the **same** marker. This means that QA can no longer test explicitly for one of the two python 3.11 containers, only for both of them.

dirkmueller:
> LGTM, but now we have a new problem: we have two different python 3.11 containers with the **same** marker. This means that QA can no longer test explicitly for one of the two python 3.11 containers, only for both of them.

I don't think thats any problem whatsoever..

alexandrevicenzi (approved):

grisu48:
> I don't think thats any problem whatsoever..

That makes test issue debugging and problem pinpointing more difficult.
If we do this, we will increase the human-based latency in the release process.

dirkmueller:
> That makes test issue debugging and problem pinpointing more difficult. If we do this, we will increase the human-based latency in the release process.

@grisu48 the container are essentially the same. I don't expect one failing and the other one succeeding. which means there is no additional overhead in test debugging.

PR #678, opened 2024-11-29
Added check for GNUtls in FIPS140 mode

pablo-herranz:
Verify that FIPS bci container is in FIPS140 mode to be compliant with Vendor Affirmation tests

https://progress.opensuse.org/issues/167683

[CI:TOXENVS] fips

### DIFF
--- a/tests/test_fips.py
+++ b/tests/test_fips.py
@@ -4,6 +4,7 @@ FIPS mode.
 """
 
 from pathlib import Path
+from re import search as regexsearch
 
 import pytest
 from pytest_container import DerivedContainer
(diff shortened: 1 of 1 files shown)


tests/test_fips.py
```diff
@@ -195,6 +195,10 @@ def test_gnutls_binary(container_per_test: ContainerData) -> None:
         "mv fips-test-gnutls /bin/fips-test-gnutls"
     )
 
+    assert "NOT" not in container_per_test.connection.check_output(
```
grisu48, line 198: This check is wrong. The requirement states:

`should return "library is in FIPS140-3 mode"or "library is in FIPS140-2 mode" or "library is in FIPS140 mode"`

So just checking for NOT is not what the requirement states.

pablo-herranz: Changed :)

tests/test_fips.py
```diff
@@ -4,6 +4,7 @@
 """
 
 from pathlib import Path
+from re import search as regexsearch
```
ricardobranco777, line 7: Why not use `re.search` which is more idiomatic?

tests/test_fips.py
```diff
@@ -189,12 +190,19 @@ def test_gnutls_binary(container_per_test: ContainerData) -> None:
...
+    c = container_per_test.connection
+
+    c.check_output(
         "zypper --gpg-auto-import-keys -n ref && zypper -n install gcc gnutls gnutls-devel && zypper -n clean && "
         "gcc -Og -g3 fips-test-gnutls.c -Wall -Wextra -Wpedantic -lgnutls -o fips-test-gnutls && "
         "mv fips-test-gnutls /bin/fips-test-gnutls"
     )
 
+    assert regexsearch(
+        r"library is in FIPS140(-3|-2|) mode",
```
ricardobranco777, line 202:
```suggestion
        r"library is in FIPS140-(2|3) mode",
```

PR #679, opened 2024-11-29
Added check for GNUtls in FIPS140 mode

dcermak:
[CI:TOXENVS] fips

CI PR for #678


tests/test_fips.py
```diff
@@ -189,14 +190,19 @@ def test_gnutls_binary(container_per_test: ContainerData) -> None:
...
         "gcc -Og -g3 fips-test-gnutls.c -Wall -Wextra -Wpedantic -lgnutls -o fips-test-gnutls && "
         "mv fips-test-gnutls /bin/fips-test-gnutls"
     )
 
-    assert "NOT" not in container_per_test.connection.check_output(
-        "gnutls-cli --fips140-mode"
+    c.check_output("gnutls-cli --fips140-mode")
+
+    assert search(
+        r"library is in FIPS140(-3|-2|) mode",
```
dirkmueller, line 204: do we really need to support the empty match, `|)` here? which distribution produces that output?

dcermak: Please add this to #678 instead

ricardobranco777 (approved):

tests/test_fips.py
dcermak: @pablo-herranz ^^^

pablo-herranz: Yes, we do as the output message might be:
- library is in FIPS140-3 mode
- library is in FIPS140-2 mode
- library is in FIPS140 mode

grisu48 (approved):

PR #680, opened 2024-12-02
Add runtime instructions

grisu48:
Adds instructions for installing dependencies and optional environment variables that might help users.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -28,11 +28,30 @@ How can I contribute?
 What do I need to contribute?
 -----------------------------
 
-* A host with python 3.6+, gcc and the python development header files
-* tox
-* go
+* A host with `python` 3.6+ and `tox`
 * docker and/or podman+buildah
 * vagrant (optional, can be used to test FIPS mode and registered hosts)
+
+Test setup on openSUSE
+^^^^^^^^^^^^^^^^^^^^^^
+
+It is recommended to run BCI tests in its own virtual environment by doing the following:
+
+.. code-block:: shell-session
+    
+    $ sudo zypper in python3-virtualenv
+    $ virtualenv .bci_tester
+    $ source .bci_tester/bin/activate
+
+Once the virtual environment `.bci_tester` has been setup, you only need to run `source .bci_tester/bin/activate` again to activate it.
+
+To setup the test environment on openSUSE Leap 15.6 and Tumbleweed, run the following commands:
+
+.. code-block:: shell-session
+
+    $ sudo zypper -n in podman buildah docker git-core python3 python3-pip
+    $ pip3 --quiet install --upgrade pip
+    $ pip3 --quiet install tox --ignore-installed six
 
 How can I run the tests?
 ------------------------
@@ -42,6 +61,27 @@ How can I run the tests?
 3. Run ``tox -e build`` (this is not strictly necessary to run beforehands, but it
    will reduce the danger of race conditions when building containers)
 4. Run ``tox -e $language_stack -- -n auto``
+
+Environment variables
+^^^^^^^^^^^^^^^^^^^^^
+
+You can set the following environment variables to configure the behavior of the BCI-Tests:
+
+.. code-block:: shell-session
+
+    $ export OS_VERSION=15.6                         # Target SLES version
+    $ export CONTAINER_RUNTIME=podman                # Defaults to podman
+    $ export TARGET=ibs-cr                           # Set container to be tested, see below
+
+Available `TARGET` settings are the following:
+
+    ibs                    SUSE:SLE-15-SP*:Update:BCI
+    ibs-cr                 Pending ToTest containers
+    ibs-released           Release BCI container
+    obs                    devel:BCI:* on OBS
+    factory-totest         Factory
+    factory-arm-totest     Factory for ARM/aarch64
+    manual                 Test the container defined by CONTAINER_URL
 
 Technical contributions
 -----------------------


README.rst
```diff
@@ -43,6 +53,17 @@ How can I run the tests?
...
+Environment variables
+^^^^^^^^^^^^^^^^^^^^^
+
+One can use the following environment variables to configure the behaviour of the BCI-Tests
+
+.. code-block:: shell-session
+
+    # export OS_VERSION=15.6                         # Target SLES version
+    # export CONTAINER_RUNTIME=podman                # Defaults to podman
+    # export TARGET=ibs-cr                           # ibs-cr = ToTest, ibs = Released, obs = Tumbleweed
```
dcermak, line 65:
```suggestion
    $ export OS_VERSION=15.6                         # Target SLES version
    $ export CONTAINER_RUNTIME=podman                # Defaults to podman
    $ export TARGET=ibs-cr                           # ibs-cr = ToTest, ibs = SUSE:SLE-15-SP*:Update:BCI, obs = devel:BCI:*
```

README.rst
```diff
@@ -28,12 +28,22 @@ How can I contribute?
 What do I need to contribute?
 -----------------------------
 
-* A host with python 3.6+, gcc and the python development header files
```
dcermak, line 31: Are you sure we don't need gcc anymore for any of the dependencies?

README.rst
```diff
@@ -28,12 +28,22 @@ How can I contribute?
...
+^^^^^^^^^^^^^^^^^^^^^^
+
+To setup the test environment on openSUSE Leap 15.6 and Tumbleweed, run the following commands:
+
+
+.. code-block:: shell-session
+
+    # zypper -n in podman buildah docker git-core python3 python3-pip
+    # pip3 --quiet install --upgrade pip
+    # pip3 --quiet install tox --ignore-installed six
```
dcermak, line 45: Let's suggest a user shell so that users don't wreck their whole OS
```suggestion
    $ sudo zypper -n in podman buildah docker git-core python3 python3-pip
    $ pip3 --quiet install --upgrade pip
    $ pip3 --quiet install tox --ignore-installed six
```

README.rst
grisu48: Yes I am, we tested this recently. See https://progress.opensuse.org/issues/168706#note-14 (Ignore `jq` and `skopeo` with are only needed in openQA)

README.rst
grisu48: Good point!

README.rst
grisu48: Thanks for the correction. I now also added `factory-totest`

README.rst
```diff
@@ -43,6 +53,17 @@ How can I run the tests?
    will reduce the danger of race conditions when building containers)
 4. Run ``tox -e $language_stack -- -n auto``
 
+Environment variables
+^^^^^^^^^^^^^^^^^^^^^
+
+One can use the following environment variables to configure the behaviour of the BCI-Tests
```
dcermak, line 59:
```suggestion
You can set the following environment variables to configure the behavior of BCI-Tests:
```

README.rst
```diff
@@ -28,12 +28,22 @@ How can I contribute?
...
+^^^^^^^^^^^^^^^^^^^^^^
+
+To setup the test environment on openSUSE Leap 15.6 and Tumbleweed, run the following commands:
+
+
+.. code-block:: shell-session
+
+    $ sudo zypper -n in podman buildah docker git-core python3 python3-pip
+    $ pip3 --quiet install --upgrade pip
+    $ pip3 --quiet install tox --ignore-installed six
```
dcermak, line 45: Maybe add a `--user` flag here?
```suggestion
    $ pip3 --quiet install --user --upgrade pip
    $ pip3 --quiet install --user tox --ignore-installed six
```

README.rst
```diff
@@ -28,12 +28,22 @@ How can I contribute?
...
+Test setup on openSUSE
+^^^^^^^^^^^^^^^^^^^^^^
+
+To setup the test environment on openSUSE Leap 15.6 and Tumbleweed, run the following commands:
+
+
+.. code-block:: shell-session
+
+    $ sudo zypper -n in podman buildah docker git-core python3 python3-pip
+    $ pip3 --quiet install --upgrade pip
```
dirkmueller, line 44: why not simply install python3-pip (or python311-pip) instead? seems simpler than this process imho.

README.rst
```diff
@@ -43,6 +53,17 @@ How can I run the tests?
...
+Environment variables
+^^^^^^^^^^^^^^^^^^^^^
+
+One can use the following environment variables to configure the behaviour of the BCI-Tests
+
+.. code-block:: shell-session
+
+    # export OS_VERSION=15.6                         # Target SLES version
+    # export CONTAINER_RUNTIME=podman                # Defaults to podman
+    # export TARGET=ibs-cr                           # ibs-cr = ToTest, ibs = SUSE:SLE-15-SP*:Update:BCI, obs = devel:BCI:*, factory-totest = Factory
```
dirkmueller, line 65: I think this is great, but we should have a section for each environment variable explaining what it means? happy to help with this myself.

README.rst
dirkmueller: I think we should advertise installation of packages, or if that is unwanted for some reason, advertise the use of venvs instead of pip install globally (or as user which is also not possible on tumbleweed)

README.rst
grisu48: But python3-pip is installed just in the line before. This is there to update pip, which we do, but I'm not sure if that's really needed.

README.rst
grisu48: I do like to advertise the use of venvs more than using the `--user` flag, so I added instructions to use a virtualenv.

By using the `--user` flag we need to explain that this will only work outside of virtualenvs and it's also not what we are using in our test flow in openQA, so I think that's not what we should document.

README.rst
grisu48: I've added a separate block now, please check again if that matches what you meant.

README.rst
dirkmueller: actually I'd prefer if we wouldn't use the pip method in openQA either from a supply chain security perspective..

README.rst
dirkmueller: It's not needed.

README.rst
```diff
@@ -28,12 +28,31 @@ How can I contribute?
...
+    $ virtualenv .bci_tester
+    $ source .bci_tester/bin/activate
+
+Once the virtual environment `.bci_tester` has been setup, you only need to run `source .bci_tester/bin/activate` again to activate it.
+
+To setup the test environment on openSUSE Leap 15.6 and Tumbleweed, run the following commands:
+
+.. code-block:: shell-session
+
+    $ sudo zypper -n in podman buildah docker git-core python3 python3-pip
```
dirkmueller, line 52: so is this supposed to be run inside the venv or outside? if it's inside the venv, you don't need to install python3-pip anymore, it is provided by the venv.

README.rst
grisu48: Let's make this a meeting point on Tuesday, that requires more more engineering work especially on older versions (12-SP5) but if we want to do this, it should be possible.

For now I suggest we decouple this initiative from this PR, because that's not easy to solve and not worth waiting for IMHO.

README.rst
grisu48: This also needs to work on older versions, so this is the most generic version that will work at least on Leap and Tumbleweed.

If we advance to specifying python311 it will not work on Leap and require us to update the documentation every time there is a Leap release or Tumbleweed updates the major python version in use.

For the sake of making the document age better, I would stay with the generic version.

README.rst
grisu48: True only if running in a venv. I phrased the previous block as optional so in cases where people run this outside a venv (for whatever reasons) it would still be needed.

README.rst
```diff
@@ -28,12 +28,31 @@ How can I contribute?
...
+
+It is recommended to run BCI tests in its own virtual environment by doing the following:
+
+.. code-block:: shell-session
+    
+    $ sudo zypper in python3-virtualenv
+    $ virtualenv .bci_tester
+    $ source .bci_tester/bin/activate
+
+Once the virtual environment `.bci_tester` has been setup, you only need to run `source .bci_tester/bin/activate` again to activate it.
```
dcermak, line 46: Tbh, I think this whole section is obsolete. I haven't activated a virtualenv in bci-tests since we introduced the `venv` toxenv.

alexandrevicenzi: True, tox should run its own virtual env AFAIK

README.rst
```diff
@@ -43,6 +62,27 @@ How can I run the tests?
...
 4. Run ``tox -e $language_stack -- -n auto``
 
+Environment variables
+^^^^^^^^^^^^^^^^^^^^^
+
+You can set the following environment variables to configure the behavior of the BCI-Tests:
+
+.. code-block:: shell-session
+
+    $ export OS_VERSION=15.6                         # Target SLES version
```
alexandrevicenzi, line 72: It would be nice to document all defaults in case nothing is set.

README.rst
dcermak: It shouldn't matter where you get `tox` from

grisu48: This virtualenv is for `tox` itself and therefore I consider it still relevant.

Especially on older SLES versions pip is the only way how to get this running and there it might be a good idea to contain it in it's own virtualenv.

I think this is still relevant.

dcermak: I have never installed pip into a virtualenv, I'd go with pipx or use pip install --user.

alexandrevicenzi: Not sure if there's anything older than py3.6 on SLE15, and in theory the tests work on that, I have no issues on Leap, but I often use venv if I want a new python version and to not install any packages on my system/user.

Most python users should be aware of poetry, venv, pipx, and so on, and might not want to use one option over another.   Perhaps it is ok to ignore this step as it could be a user preference.

grisu48: This is the way how the test run currently in openQA. It is a help to the user and especially newcomers.

README.rst
grisu48: Let's address this in another PR.

grisu48:
@dcermak @dirkmueller @alexandrevicenzi I see still unresolved comments waiting to be resolved or commented on. Can you PTAL?

README.rst
dcermak: You install tox into a virtualenv and then launch tox from that?

grisu48: Yes, see e.g. https://github.com/os-autoinst/os-autoinst-distri-opensuse/blob/master/tests/containers/bci_prepare.pm#L130

That's the only way how we could handle this on all requested systems of the test matrix, SLES and non-SLES alike.

dcermak: That's just… wrong. Sorry to be so harsh, but that's neither a suggested method by upstream: https://tox.wiki/en/latest/installation.html nor something that I would personally recommend doing.

I see that you have to jump through hoops for openQA, but I don't want to be recommending that for regular developers.

grisu48: I understand the hesitation, but Testing depends on this workflow and it should be also documented somewhere. I see only two options here: Either proceed and document it in the BCI-tests themselves for the purpose of making it easier for testes to get started, or we keep it downstream.

The former has the advantage that relevant documentation stays with the project itself but we rely on discouraged methods (which we anyways do) and avoids further fragmentation of the knowledge about this project.

That is the whole reason for the PR.

@dcermak take a moment to consider this point of view as well. I do understand the concerns, but in the context of where this project lives, I find the argument of not relying on yet-another-Confuence-page more important than trying to rely on `tox` ideoms which we cannot fulfill on the whole test matrix. It's a compromise either way.

README.rst
dirkmueller: Given that tox sets up the virtualenvs on its own, why do we need to do that?

README.rst
dirkmueller: We could also simply be fixing the testing setup instead? there's no good reason for not using the distribution provided tox (other than that the 3.6 tox on SLE15 is broken, but there is a python311-tox which works and which openQA is afaik using atm).

dcermak: @grisu48 I see your point, but I definitely don't want a discouraged installation/setup method at the top of the documentation. If openQA has to do it in such a quirky way, then so be it, but then it should **not** be at the top of the documentation, but instead hidden under a spoiler warning saying "only open if all hope is lost"

README.rst
grisu48: Because `tox` is not available in the repositories of all distribution we are supposed to run this on.

We need to get it from pip.

grisu48:
Closing as of https://github.com/SUSE/BCI-tests/pull/680#discussion_r1949020247.

PR #681, opened 2024-12-03
Add xfail for the openjdk-devel on ppc64le issue

dirkmueller:

### DIFF
--- a/tests/test_openjdk_devel.py
+++ b/tests/test_openjdk_devel.py
@@ -7,6 +7,7 @@ import pytest
 from pytest_container import DerivedContainer
 from pytest_container import container_and_marks_from_pytest_param
 from pytest_container.container import ContainerData
+from pytest_container.runtime import LOCALHOST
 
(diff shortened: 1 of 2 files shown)


dcermak (dismissed):

tests/test_openjdk_devel.py
```diff
@@ -86,6 +87,10 @@ def test_maven_present(auto_container):
     assert auto_container.connection.run_expect([0], "mvn --version")
 
 
+@pytest.mark.xfail(
```
grisu48, line 90: This is good but not enough. See https://openqa.suse.de/tests/16039106, we need to softfail also the following test runs:

* `test_glibc_present`
* `test_product`
* `test_bci_eula_is_correctly_available`
* `test_java_home`
* `test_maven_present`

All of them fail due to the "can only create exec sessions on running containers: container state improper" error because the container cannot be started.

dirkmueller: okay, we can fix that by setting a custom entry point in the tests

grisu48 (approved):

PR #688, opened 2024-12-05
Add smoke tests for stunnel

dcermak:
[CI:TOXENVS] stunnel

for https://github.com/SUSE/BCI-dockerfile-generator/pull/2110


alexandrevicenzi (commented):
Overall LGTM, but it needs a rebase.

tests/test_stunnel.py
```diff
@@ -0,0 +1,179 @@
...
+def _create_http_forward_ctr(log_level: str = "info") -> ParameterSet:
+    return pytest.param(
+        DerivedContainer(
+            **_stunnel_kwargs,
+            forwarded_ports=[PortForwarding(container_port=_CTR_PORT)],
+            extra_environment_variables={
+                "STUNNEL_DEBUG": log_level,
+                "STUNNEL_SERVICE_NAME": "wikipedia_dot_org",
+                "STUNNEL_ACCEPT": f"0.0.0.0:{_CTR_PORT}",
+                "STUNNEL_CONNECT": "wikipedia.org:80",
```
grisu48, line 125: I would recommend to use some of our own resources, e.g. conncheck.opensuse.org or something similar and avoid using wikipedia for reasons other than it's actual purpose. Even when it's not directly used now, it could be used by accident.

Otherwise LGTM.
```suggestion
                "STUNNEL_CONNECT": "conncheck.opensuse.org:80",
```

tests/test_stunnel.py
```diff
@@ -0,0 +1,180 @@
...
+
+Stunnel requires a certificate to create TLS connections. We create a self
+signed certificate for that in the :file:`tests/files/stunnel/` directory as
+follows:
+
+.. code-block:: shell-session
+
+   $ openssl genrsa -out server.key 2048
+   $ openssl req -new -key server.key -out server.csr
+   $ openssl x509 -req -days 3650 -in server.csr -signkey server.key -out server.crt -extfile v3.ext
```
alexandrevicenzi, line 11: The certificate is going to expire one day, I'm not sure if there's a way to document that, otherwise if this code lives for 10+ years you get a failure to debug.

tests/test_stunnel.py
```diff
@@ -0,0 +1,180 @@
...
+def _create_http_forward_ctr(log_level: str = "info") -> ParameterSet:
+    return pytest.param(
+        DerivedContainer(
+            **_stunnel_kwargs,
+            forwarded_ports=[PortForwarding(container_port=_CTR_PORT)],
+            extra_environment_variables={
+                "STUNNEL_DEBUG": log_level,
+                "STUNNEL_SERVICE_NAME": "neverssl",
+                "STUNNEL_ACCEPT": f"0.0.0.0:{_CTR_PORT}",
+                "STUNNEL_CONNECT": "neverssl.com:80",
```
alexandrevicenzi, line 125: can we trust this won't cause random failures in CI?

alexandrevicenzi (dismissed):

tests/test_stunnel.py
dcermak: I've added a warning

tests/test_stunnel.py
```diff
@@ -0,0 +1,188 @@
...
+            verify=_SERVER_PEM,
+            timeout=5,
+        )
+        resp.raise_for_status()
+        return resp
+
+    resp = get_request_to_stunnel()
+    assert resp.status_code == 200
+    assert (
+        "This website is for when you try to open Facebook, Google, Amazon, etc"
```
grisu48, line 162:
```suggestion
        assert "neverssl" in resp.text.lower()
```
Making assumptions on external resources we have no control over is always difficult. I would limit the assumption to a minimum, so that the test doesn't fail when someone slightly changes the text on neverssl.com. We already check for status code 200, I think this check doesn't need to be extensive here.

grisu48 (dismissed):
Small nitpick, otherwise LGTM.

alexandrevicenzi (dismissed):

dirkmueller (approved):

PR #691, opened 2024-12-10
Update to more stable tag for kiwi on tumbleweed

dirkmueller:
It recently updated from 10.1 to 10.2, we can just list `:latest` which is available and pointing to the right container.


alexandrevicenzi (dismissed):

dcermak (changes requested):
`'kiwi_10' not found in `markers` configuration option`

dcermak (approved):

PR #692, opened 2024-12-10
Don't use removesuffix, use a regex instead

dcermak:
str.removesuffix requires python 3.9+, but we must stay 3.6 compatible

[CI:TOXENVS] metadata

### DIFF
--- a/tests/test_metadata.py
+++ b/tests/test_metadata.py
@@ -14,6 +14,7 @@ offer the labels under the ``com.suse.bci`` prefix but ``com.suse.sle``.
 
 """
 
+import re
 import urllib.parse
 from pathlib import Path
 from typing import List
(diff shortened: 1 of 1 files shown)


tests/test_metadata.py
```diff
@@ -385,7 +386,7 @@ def test_general_labels(
         # no EULA for openSUSE images
     else:
         assert (
-            labels["com.suse.lifecycle-url"].removesuffix("/")
+            re.sub(r"\/$", "", labels["com.suse.lifecycle-url"])
```
alexandrevicenzi, line 389: nit: compared to the original version, `re.sub` has the potential to remove in any place, while `removesuffix` only removes if it matches the suffix.

alexandrevicenzi (approved):

tests/test_metadata.py
dcermak: Nope, that's what the `$` (= end of string) is for:
```
>>> re.sub(r'\/$', '', "foo.bar.com/bar/fooo/")
'foo.bar.com/bar/fooo'
```

dirkmueller: can we simply list the erratic url in the tuple below? that way we don't allow a trailing slash for every of the option which might be incorrect. 

or we could simply fix the containers..

dirkmueller: I've fixed the containers now, so the whole shebang can be dropped.

alexandrevicenzi: I missed the `$`, and I agree with @dirkmueller, better to fix the source issue.

dirkmueller:
no longer needed

PR #693, opened 2024-12-10
Avoid removesuffix()

dirkmueller:
This function was only introduced in python 3.9, but QA tests with python 3.6.


alexandrevicenzi:
Dup of #692?

dirkmueller:
Yeah, didn't see that in time.

alexandrevicenzi (approved):

PR #696, opened 2024-12-11
Limit softfailure for bsc#1221983 to JDK21

grisu48:
The jshell failure is only present on OpenJDK-devel 21, so we need to limit the softfailure also to just this Java version. See https://bugzilla.suse.com/show_bug.cgi?id=1221983#c1

[CI:TOXENVS] openjdk_devel

### Verification runs:

* [openjdk-devel-11](https://openqa.suse.de/tests/16183529#step/_root_BCI-tests_openjdk_/4)
* [openjdk-devel-17](https://openqa.suse.de/tests/16183530#step/_root_BCI-tests_openjdk_/4)
* [openjdk-devel-21](https://openqa.suse.de/tests/16183531#step/_root_BCI-tests_openjdk_/4)

### DIFF
--- a/tests/test_openjdk_devel.py
+++ b/tests/test_openjdk_devel.py
@@ -87,13 +87,23 @@ def test_maven_present(auto_container):
     assert auto_container.connection.run_expect([0], "mvn --version")
 
 
-@pytest.mark.xfail(
-    LOCALHOST.system_info.arch == "ppc64le",
-    reason="https://bugzilla.suse.com/show_bug.cgi?id=1221983",
-)
 @pytest.mark.parametrize(
     "container,java_version",
-    CONTAINER_IMAGES_WITH_VERSION,
+    # Softfailure for bsc#1221983 only applies to OpenJDK-devel-21
+    [
+        param
+        if param.values[1] != 21
+        else pytest.param(
+            *param.values,
+            id=param.id,
+            marks=param.marks
+            + pytest.mark.xfail(
+                condition=LOCALHOST.system_info.arch == "ppc64le",
+                reason="https://bugzilla.suse.com/show_bug.cgi?id=1221983",
+            ),
+        )
+        for param in CONTAINER_IMAGES_WITH_VERSION
+    ],
     indirect=["container"],
 )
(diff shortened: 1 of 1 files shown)


dirkmueller:
@grisu48 please run tox -e format

grisu48:
> @grisu48 please run tox -e format

I'm stuck elsewhere, please see https://suse.slack.com/archives/C02AF8LALDA/p1733901150518579?thread_ts=1733900588.468239&cid=C02AF8LALDA

tests/test_openjdk_devel.py
```diff
@@ -87,13 +87,13 @@ def test_maven_present(auto_container):
...
-)
 @pytest.mark.parametrize(
     "container,java_version",
     CONTAINER_IMAGES_WITH_VERSION,
+    marks=pytest.mark.xfail(
+        condition=LOCALHOST.system_info.arch == "ppc64le" and java_version == 11,
+        reason="https://bugzilla.suse.com/show_bug.cgi?id=1221983",
+    ),
     indirect=["container"],
 )
```
dcermak, line 98: `java_version` is not available in this scope, what you can do is roughly the following:
```suggestion
@pytest.mark.parametrize(
    "container,java_version",
    [param for param in CONTAINER_IMAGES_WITH_VERSION if param.values[1] != 11 else pytest.param(*param.values, id=param.id, marks=param.marks + pytest.mark.xfail(
        condition=LOCALHOST.system_info.arch == "ppc64le",
        reason="https://bugzilla.suse.com/show_bug.cgi?id=1221983",
    ),
    indirect=["container"],
)
```

dirkmueller:
> The jshell failure is only present on OpenJDK-devel 11, so we need to limit the softfailure also to just this Java version.

can you point where the xfail is passing?

according to my emulated testing this isn't limited to openjdk 11 at all, and the referenced bugreport confirms it happens with 21.

tests/test_openjdk_devel.py
grisu48: Thanks for the hint! Should be fixed now.

dirkmueller (approved):

PR #699, opened 2024-12-11
[tomcat] stop needing curl for the tests

dcermak:
[CI:TOXENVS] tomcat

### DIFF
--- a/tests/test_tomcat.py
+++ b/tests/test_tomcat.py
@@ -11,8 +11,7 @@ from pytest_container.container import ContainerData
 from bci_tester.data import TOMCAT_CONTAINERS
 
 
-@pytest.mark.parametrize("container", TOMCAT_CONTAINERS, indirect=True)
-def test_tomcat_launches(container: ContainerData) -> None:
+def _tomcat_launch_test_fn(container: ContainerData) -> requests.Response:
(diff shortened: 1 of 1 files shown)


tests/test_tomcat.py
```diff
@@ -47,29 +50,91 @@ def _fetch_tomcat_root() -> requests.Response:
...
-        f"http://localhost:{container_per_test.forwarded_ports[0].host_port}/sample",
-        timeout=2,
+
+    mount = container_per_test.container.volume_mounts[0]
+    assert isinstance(mount, BindMount) and mount.host_path
+
+    # make the tmpdir world readable + writeable, otherwise the tomcat container
+    # won't be able to read it, as it runs under the tomcat user which is mapped
+    # to some random subuserid that has no permissions
+    os.chmod(mount.host_path, 0o0777)
```
dirkmueller, line 109: why can't we just simply " podman cp" it into the running container like we do in all other tests? we don't need to setup all the bind mount stuff. 

I had a half finished patch for that already somewhere.

dcermak: > I had a half finished patch for that already somewhere.

Perfect is the enemy of done. Nevertheless, here you go

dirkmueller (approved):
Thanks!

PR #701, opened 2024-12-11
Fix repository substitution for Tumbleweed

dirkmueller:
There is no perl anymore, and we also didn't check that the substitution actually happened.

### DIFF
--- a/bci_tester/data.py
+++ b/bci_tester/data.py
@@ -220,15 +220,22 @@ BCI_DEVEL_REPO = os.getenv("BCI_DEVEL_REPO")
 
 
 if BCI_DEVEL_REPO is None:
-    # from SLE 15 SP6 onward we use the unauthenticated CDN
-    cdn_prefix = "installer-updates" if OS_SP_VERSION >= 6 else "updates"
-    BCI_DEVEL_REPO = f"https://{cdn_prefix}.suse.com/SUSE/Products/SLE-BCI/{OS_MAJOR_VERSION}-SP{OS_SP_VERSION}/{LOCALHOST.system_info.arch}/product/"
+    if OS_VERSION == "tumbleweed":
+        BCI_DEVEL_REPO = "http://download.opensuse.org/tumbleweed/repo/oss/"
+    else:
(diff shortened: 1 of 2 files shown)


bci_tester/data.py
```diff
@@ -220,15 +220,22 @@ def create_container_version_mark(
...
+        cdn_prefix = "installer-updates" if OS_SP_VERSION >= 6 else "updates"
+        BCI_DEVEL_REPO = f"https://{cdn_prefix}.suse.com/SUSE/Products/SLE-BCI/{OS_MAJOR_VERSION}-SP{OS_SP_VERSION}/{LOCALHOST.system_info.arch}/product/"
     _BCI_REPLACE_REPO_CONTAINERFILE = ""
 else:
     bci_repo_path = f"/etc/zypp/repos.d/{BCI_REPO_NAME}.repo"
-    # only try to replace the baseurl if the file exists and we have perl
-    # available (should be more ubiquitous than sed)
-    _BCI_REPLACE_REPO_CONTAINERFILE = f"RUN if [ -e {bci_repo_path} ] && [ -e /usr/bin/perl ]; then perl -pi -e 's|baseurl.*|baseurl={BCI_DEVEL_REPO}|g' {bci_repo_path}; fi"
+    bci_repo_replace = "${line/baseurl*/baseurl = " + BCI_DEVEL_REPO + "}"
+    _BCI_REPLACE_REPO_CONTAINERFILE = f"""RUN if [ -e {bci_repo_path} ]; then \\
```
grisu48, line 233: I don't get why we allow for multiple repositories here, but assert that there is only one in `test_all.py` below.

dirkmueller: There are always multiple repos. In test_all we are testing that we can find exactly one with the expected name because that's the one we're going to modify

grisu48 (approved):

grisu48:
I will run a bunch of test runs this morning, and then we can merge.

grisu48:
VR passing: https://openqa.opensuse.org/tests/4699165, let's merge.

PR #711, opened 2024-12-20
Skip tests that depend on the BCI repo for the LTSS container

dirkmueller:
We have the LTSS container on a service pack still in general support, so a container-specific exclusion rather than merely a version based exclusion is needed.

### DIFF
--- a/tests/test_all.py
+++ b/tests/test_all.py
@@ -33,6 +33,7 @@ from bci_tester.data import DISTRIBUTION_CONTAINER
 from bci_tester.data import INIT_CONTAINER
 from bci_tester.data import KERNEL_MODULE_CONTAINER
 from bci_tester.data import KIWI_CONTAINERS
+from bci_tester.data import LTSS_BASE_CONTAINERS
 from bci_tester.data import MICRO_CONTAINER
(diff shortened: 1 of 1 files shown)


dcermak (commented):
LGTM in general, but  I think handling this in a decorator would be more obvious as that's where we generally handle skipifs

dirkmueller:
> LGTM in general, but I think handling this in a decorator would be more obvious as that's where we generally handle skipifs

it is a lot of repetition imho. anyhow, updated.

m-dati:
LGTM, but some CI checks fail

dcermak (approved):

dcermak (changes requested):
Well, we also need to fix the remaining test (metadata for instance is red as well)

dirkmueller:
> Well, we also need to fix the remaining test (metadata for instance is red as well)

did you check the error?

> Command '['docker', 'pull', 'registry.suse.com/suse/ltss/sle15.5/bci-base:15.5']' returned non-zero exit status 1

That's because this PR is not being merged. metadata checks are green in openqa, the only thing blocking it from being erleased is this PR.

dirkmueller:
moving ahead, we need to get the container out..

PR #715, opened 2025-01-07
[openjdk] explicitly set entrypoint to bash

dcermak:
this prevents the containers from launching jshell potentially interfering with the tests, see [poo#174976](https://progress.opensuse.org/issues/174976).

[CI:TOXENVS] openjdk_devel


ricardobranco777 (approved):
it works: https://openqa.suse.de/tests/16363935

PR #716, opened 2025-01-08
Switch SP6+ to public-dl.suse.com

dirkmueller:
We'll move installer-updates to a fixed-ip setup to maximize interoperability. However BCI should use the direct CDN link, so switch to that.


alexandrevicenzi (approved):

dcermak:
Please hold this until the maintenance updates go through. Otherwise we'll cause a nice chicken-or-err problem for QA again

dirkmueller:
now the maintenance update is stuck in QA due to this missing.

PR #724, opened 2025-01-13
[ruby] Switch Ruby to the latest tag for TW

dcermak:
With this, we no longer have to switch the build tag on every new release
[CI:TOXENVS] ruby


grisu48 (approved):
Thank you!

bci_tester/data.py
```diff
@@ -617,7 +617,7 @@ def create_BCI(
 )
 
 RUBY_34_CONTAINER = create_BCI(
```
grisu48, line 619:
```suggestion
RUBY_CONTAINER = create_BCI(
```

dcermak: better?

dirkmueller (approved):

PR #726, opened 2025-01-16
Simplify test code given the changes from pytest_container#216

dcermak:
We now no longer have to perform the manual container & mark unbundling as `Container` and `DerivedContainer` are now instances of `ParameterSet` as well

See also https://github.com/dcermak/pytest_container/pull/216


alexandrevicenzi (commented):
Overall, looks a lot cleaner :+1:

dirkmueller (approved):

PR #734, opened 2025-02-03
Add a pre-commit hook to prevent most common mistakes

dcermak:
Don't you hate the CI failing because you forgot to run `tox -e format`? Or one of the CI jobs starting to fail because you didn't add a marker to `pyproject.toml`?

Now you can get the failure on trying to commit :joy:!


.github/workflows/ci.yaml
```diff
@@ -22,9 +22,9 @@ jobs:
         run: |
           set -e
           sudo apt update
-          sudo apt -y install jo tox
-      - name: Test formatting with ruff
-        run: tox -e format -- --check
```
grisu48, line 27: Why not both? I mean, people do weird things on their computers all the time, I would not rely on local tests.

Otherwise an annoyed engineer that decides to bypass the pre-commit can poison the repo for everyone else.

bci_tester/data.py
```diff
@@ -1137,11 +1138,19 @@ def create_BCI(
 
 
 if __name__ == "__main__":
-    import json
+    from typing import Set
```
grisu48, line 1141: Maybe it's just my tired brain, but why are those changes needed for this PR? :thinking:

.github/workflows/ci.yaml
dcermak: check the lines below, they run the pre-commit script, which executes `tox -e format -- --check`

bci_tester/data.py
dcermak: This is the actual code change that checks whether all markers are present in `pyproject.toml`. Before that, it would print the container base urls, but this code path hasn't been used in ages

.github/workflows/ci.yaml
grisu48: I should not review PRs when I'm tired. Sorry, of course.

grisu48 (approved):

PR #736, opened 2025-02-04
Don't pass in private key for (public) certificate pinning

dirkmueller:
No idea what was intended with this, but it certainly isn't correct..

[CI:TOXENVS] stunnel


grisu48 (approved):
Yes, this should work.

grisu48 (dismissed):

tests/test_stunnel.py
```diff
@@ -106,7 +106,7 @@ def test_stunnel_http_proxy(pod: PodData):
     def get_request_to_stunnel(route: str = "") -> requests.Response:
         resp = requests.get(
             f"https://127.0.0.1:{pod.forwarded_ports[0].host_port}/{route}",
-            verify=_SERVER_PEM,
+            verify=_SERVER_CRT if sys.version_info >= (3, 7) else False,
```
dcermak, line 109: This needs an explanation why please

dirkmueller: I don't have a why. all I know is that it breaks with 1.1.1d and works with 1.1.1l or newer. there is 550000 lines of differences between then, as it works with any newer version of openssl I think it's just a bug and I don't have time to do reverse archeology on some old openssl version that we can not update anymore as it is not maintained anymore.

dcermak (approved):

tests/test_stunnel.py
dcermak: That is more than enough, I never expected the full breakdown, just a note explaining why this condition is there. What you added is fully sufficient, thanks!

PR #737, opened 2025-02-04
Test that ca-certificates are working in the base container

dirkmueller:


dcermak (commented):
But we already have this: https://github.com/SUSE/BCI-tests/blob/893aeb9cd27c130005eb9ce85ae1a099459a9552/tests/test_all.py#L529

dirkmueller:
> But we already have this:
> 
> https://github.com/SUSE/BCI-tests/blob/893aeb9cd27c130005eb9ce85ae1a099459a9552/tests/test_all.py#L529

I think it didn't trigger... will retest

PR #739, opened 2025-02-06
Use a webserver for tests that is under our control (bsc#1236833)

dirkmueller:
[CI:TOXENVS] python

### DIFF
--- a/tests/test_python.py
+++ b/tests/test_python.py
@@ -269,8 +269,8 @@ def test_python_webserver_1(
         (
             BCDIR + OUTDIR,
             "communication_examples.py",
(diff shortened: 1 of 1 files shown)


alexandrevicenzi (commented):
Python wget last release seems to be 2015 and the project in bitbucket is gone.

I think it would be worth to convert the code to requests, WDYT? Should we create another ticket to rewrite that price of code?

alexandrevicenzi (approved):

dirkmueller:
> I think it would be worth to convert the code to requests, WDYT? Should we create another ticket to rewrite that price of code?

I don't know why this module was chosen in that test, but I agree it doesn't look like the best testcase. I think this could be done in another PR.

PR #740, opened 2025-02-06
Refactor python tests

dcermak:
[CI:TOXENVS] python


tests/test_python.py
```diff
@@ -207,113 +216,37 @@ def test_pip_install_source_cryptography(auto_container_per_test):
...
 
-    # check expected file not present yet in the destination
-    assert not container_per_test.connection.file(destdir + xfilename).exists
-
-    # execution of the python module in the container
-    bci_python_wget = container_per_test.connection.check_output(
-        f"timeout --preserve-status 120s python3 {APPDIR + appl2} {url} {destdir}",
+    """
+    LOGO_URL = "https://opensource.suse.com/bci/SLE_BCI_logomark_green.svg"
+    assert (
```
alexandrevicenzi, line 243: nit: if the request fails, e.g, returns a 404, it will still work because only the text is checked, not if the request succeeded.

Consider a `raise_for_status` to ensure we get a proper 2xx response.

alexandrevicenzi (dismissed):

tests/test_python.py
dcermak: Good catch, I've added a `raise_for_status` in both places

tests/test_python.py
```diff
@@ -207,113 +216,42 @@ def test_pip_install_source_cryptography(auto_container_per_test):
...
+    _LOGO_URL = "https://opensource.suse.com/bci/SLE_BCI_logomark_green.svg"
+    logo_resp = requests.get(_LOGO_URL)
+    logo_resp.raise_for_status()
 
-    # execution of the python module in the container
-    bci_python_wget = container_per_test.connection.check_output(
-        f"timeout --preserve-status 120s python3 {APPDIR + appl2} {url} {destdir}",
+    assert (
+        container_per_test.connection.check_output(
+            f"source .venv/bin/activate; python -c \"import requests; print(requests.get('{_LOGO_URL}').text)\""
```
dirkmueller, line 250: The timeout=120 from previous code is lost now, isn't it?

tests/test_python.py
```diff
@@ -207,113 +216,42 @@ def test_pip_install_source_cryptography(auto_container_per_test):
...
+        )
+        == logo_resp.text
```
dirkmueller, line 252: This fetches the logo twice per test.. we could avoid half of it by comparing against a known checksum for example.

tests/test_python.py
dcermak: I've added it back

tests/test_python.py
dcermak: I've changed it to compare the sha512sum

dirkmueller (approved):
Thanks!

PR #752, opened 2025-02-18
Adding tests for AI pytorch container

chajain:

### DIFF
--- a/bci_tester/data.py
+++ b/bci_tester/data.py
@@ -1029,6 +1029,13 @@ MILVUS_CONTAINER = create_BCI(
     custom_entry_point="/bin/bash",
 )
 
+PYTORCH_CONTAINER = create_BCI(
+    build_tag=f"{SAC_CONTAINER_PREFIX}/pytorch:2.5.0",
+    bci_type=ImageType.SAC_APPLICATION,
+    available_versions=["15.6-ai"],
+    custom_entry_point="/bin/bash",
+)
+
 STUNNEL_CONTAINER = create_BCI(
     build_tag=f"{APP_CONTAINER_PREFIX}/stunnel:5",
     bci_type=ImageType.APPLICATION,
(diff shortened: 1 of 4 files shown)


tests/test_ai.py
```diff
@@ -71,3 +77,18 @@ def test_milvus_health(container_per_test):
...
+    indirect=["container_per_test"],
+)
+def test_pytorch_health(container_per_test):
+    """Test the pytorch container."""
+
+    # chech pytorch Version
+    container_per_test.connection.check_output(
+        "python3.11 -c 'import torch; print(torch.__version__)'"
+    )
+    container_per_test.connection.check_output("git --version")
```
dcermak, line 94: No need to use the per_test variant here
```suggestion
    "container",
    [PYTORCH_CONTAINER],
    indirect=["container"],
)
def test_pytorch_health(container):
    """Test the pytorch container."""

    # chech pytorch Version
    container.connection.check_output(
        "python3.11 -c 'import torch; print(torch.__version__)'"
    )
    container.connection.check_output("git --version")
```

dcermak (commented):
A bit bare bones but should work

guangyee:
@chajain looks like we should be using the `2.5.0` tag instead of `2.5`. See registry.suse.de/suse/sle-15-sp6/update/products/ai/totest/images/containers/pytorch:2.5.0

dirkmueller (approved):

bci_tester/data.py
```diff
@@ -1029,6 +1029,13 @@ def create_BCI(
     custom_entry_point="/bin/bash",
 )
 
+PYTORCH_CONTAINER = create_BCI(
+    build_tag=f"{SAC_CONTAINER_PREFIX}/pytorch:2.5.0",
```
dcermak, line 1033: You will have to bump this string on every version bump in pytorch, which seems extremely wasteful to me. Can you use a more generic tag? Or are you really expecting to stay at 2.5.0 for an extended period of time?

dcermak:
Closing in favor of #768

PR #754, opened 2025-02-19
Test that opencl-devel is available (jsc#PED-7838)

dirkmueller:


dcermak:
That still fails on SP7, have you already submitted a fix?

dirkmueller:
> That still fails on SP7, have you already submitted a fix?

now even with a submitrequest ;)

PR #758, opened 2025-02-25
drop kubernetes 1.29 from Tumbleweed

dirkmueller:
[CI:TOXENVS] kubectl


dcermak:
Can you also please drop the kubectl 1.29 marker from pyproject.toml?

dirkmueller:
> Can you also please drop the kubectl 1.29 marker from pyproject.toml?

done

PR #765, opened 2025-03-12
Add maintainer tag

grisu48:
Adds a maintainer/owner tag to each test module that helps to identify the test/container owners.

### DIFF
--- a/tests/test_389ds.py
+++ b/tests/test_389ds.py
@@ -1,5 +1,7 @@
 """Basic tests for the 389-ds Application container image."""
 
+## Maintainer: BCI team (#proj-bci)
+
 from pytest_container.runtime import LOCALHOST
 
 from bci_tester.data import CONTAINER_389DS_CONTAINERS
--- a/tests/test_ai.py
+++ b/tests/test_ai.py
@@ -1,5 +1,7 @@
 """This module contains the tests for the SUSE AI container."""
 
+## Maintainer: AI team (#help-ai-technical)
+
 import pytest
 import requests
 from tenacity import retry
--- a/tests/test_all.py
+++ b/tests/test_all.py
@@ -2,6 +2,8 @@
 This module contains tests that are run for **all** containers.
 """
 
+## Maintainer: BCI team (#proj-bci)
+
 import datetime
 import fnmatch
 import json
--- a/tests/test_base.py
+++ b/tests/test_base.py
@@ -3,6 +3,8 @@ registry.suse.com)
 
 """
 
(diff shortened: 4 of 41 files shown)


dcermak (commented):
Do we really need this? I mean this looks to me more like that the AI team should get their own testsuite, given that the AI containers don't even play nicely with this one at all.

grisu48:
> Do we really need this? I mean this looks to me more like that the AI team should get their own testsuite, given that the AI containers don't even play nicely with this one at all.

I'm also happy with this solution.

On [Tueday](https://confluence.suse.com/display/ENGCTNRSTORY/2025-03-11+-+BCI+Sync) I raised the topic that it's difficult for QE to find the responsible person for a container, and this was an agreed solution for this issue.

If we agree that SUSE AI will have their own testsuite, it would not be necessary.

dirkmueller:
There was a separate project to create testing coverage for SUSE AI. this testsuite is not used to validate the container source changes, it is used to validate the rebuilds due to dependency changes. as such it doesn't need to "play nicely" just figure out if the container are still starting.

dirkmueller:
> Adds a maintainer/owner tag to each test module that helps to identify the test/container owners.

@grisu48 this is confusing to me, I thought we agreed on adding a maintainer tag in https://gitlab.suse.de/qac/container-release-bot/-/tree/main/containers?ref_type=heads instead where all the testing gets in one place ? why would you want to have that in BCI-tests when not all containers that the container-release-bot/openqa is handling is going via BCI-tests?

dcermak:
> There was a separate project to create testing coverage for SUSE AI. this testsuite is not used to validate the container source changes, it is used to validate the rebuilds due to dependency changes. as such it doesn't need to "play nicely" just figure out if the container are still starting.

It does not play nicely insofar that it doesn't integrate into our development workflow at all. Currently we cannot test that we don't accidentally break the AI tests with a source change. It hasn't happened yet, but I'm not too thrilled about that eventually happening. Also it makes testing AI related contributions like https://github.com/SUSE/BCI-tests/pull/752 a major pain.

dirkmueller:
> It does not play nicely insofar that it doesn't integrate into our development workflow at all. Currently we cannot test that we don't accidentally break the AI tests with a source change.

This is the same for anything else that is built in IBS, like the LTSS containers, and the same fix can be applied (running the testsuite against TARGET=ibs-released so that we can at least validate the test code still working fine. It wasn't possible when the code was initially introduced due to chicken-and-egg situation but now this can be easily done.

PR #767, opened 2025-03-13
Redirect testinfra logging to a per worker logfile

dcermak:
See: https://progress.opensuse.org/issues/174634

[CI:TOXENVS] base

Currently this papertrail is created on every test run. I think we could gate this via a CLI flag or an environment variable. wdyt @ricardobranco777


ricardobranco777 (approved):
LGTM. Thanks!

ricardobranco777 (approved):

PR #770, opened 2025-03-18
Drop SUSE AppCollection specific containers from testing

dirkmueller:
Keeping the test infrastructure in tact in case it gets revisited.

### DIFF
--- a/bci_tester/data.py
+++ b/bci_tester/data.py
@@ -603,16 +603,6 @@ PYTHON_CONTAINERS = PYTHON_WITH_PIPX_CONTAINERS + [
     )
 ]
 
-# Python containers on SUSE Application Collection
-SAC_PYTHON_CONTAINERS = [
-    create_BCI(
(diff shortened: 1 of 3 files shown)


dcermak (commented):
I think we can also stop marking postfix & tomcat as `ImageType.SAC_APPLICATION` in lines 181 and 217 and instead use `ImageType.Application`

dirkmueller:
> I think we can also stop marking postfix & tomcat as `ImageType.SAC_APPLICATION` in lines 181 and 217 and instead use `ImageType.Application`

done

dcermak:
> > I think we can also stop marking postfix & tomcat as `ImageType.SAC_APPLICATION` in lines 181 and 217 and instead use `ImageType.Application`
> 
> done

Should've specified the file :sweat_smile: , I thought about `test_metadata.py`

dirkmueller:
> Should've specified the file 😅 , I thought about `test_metadata.py`

done as well.

dcermak (approved):

PR #773, opened 2025-03-27
Drop sha1 from the FIPS digests for SLE 16

dcermak:
https://jira.suse.com/browse/PED-12198

[CI:TOXENVS] all


denknorr:
Yes, please remove SHA1. SHA1 is not considered as approved in FIPS mode and has not to be used for anything past 2030 anyway. And it's known to be broken for years.

Thank you.

PR #778, opened 2025-04-03
Initial testing for PrivateRegistry

dirkmueller:

### DIFF
--- a/bci_tester/data.py
+++ b/bci_tester/data.py
@@ -46,6 +46,7 @@ ALLOWED_BASE_OS_VERSIONS = (
 ALLOWED_NONBASE_OS_VERSIONS = (
     "15.6",
     "15.6-ai",
+    "15.6-pr",
     "15.7",
     "16.0",
     "tumbleweed",
@@ -55,6 +56,7 @@ ALLOWED_NONBASE_OS_VERSIONS = (
 ALLOWED_BCI_REPO_OS_VERSIONS = (
     "15.6",
     "15.6-ai",
+    "15.6-pr",
     "15.7",
     "tumbleweed",
 )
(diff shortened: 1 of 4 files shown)


bci_tester/data.py
```diff
@@ -1018,14 +1035,20 @@ def create_BCI(
         forwarded_ports=[PortForwarding(container_port=6379)],
     )
     for versions, tag in ((("tumbleweed", "15.6", "15.7"), "8.0"),)
+] + [
+    create_BCI(
+        build_tag="private-registry/harbor-valkey:8.0.2",
```
dcermak, line 1040: Isn't there a tag that is not so specific? This is going to require a lot of churn :-(

bci_tester/data.py
```diff
@@ -174,6 +176,11 @@
             "registry.suse.de/suse/sle-15-sp6/update/products/ai/totest"
         )
 
+    if OS_VERSION == "15.6-pr":
+        ibs_cr_project = (
+            "registry.suse.de/suse/sle-15-sp6/update/products/privateregistry"
```
dcermak, line 181: Another project that cannot be tested on github?

bci_tester/data.py
dirkmueller: there's a latest tag also

bci_tester/data.py
dirkmueller: it can be tested on github via ibs-released mode, which we can enable once we actually managed to release it.

bci_tester/data.py
dcermak: let's use that please

bci_tester/data.py
dcermak: We'd have to ensure that this is doesn't default to ibs-released on openQA, but otherwise I'd be very much in favor

alexandrevicenzi (approved):

dirkmueller:
> Given that the SUSE AI team is reconsidering the usage of BCI-Tests, does it make sense to add another suse internal project here, without any meaningful test coverage?

two different teams with two entirely different goals (and delivery pipelines?). I think we should have that discussion elsewhere..

PR #786, opened 2025-04-15
Add a smoke test that all users in the container have a stable uid & gid

dcermak:
First draft to test whether uids & gids are stable. Currently this _should_ pass and establishes a baseline that we can rely on, but has some limitations:
- the entrypoints of pcp & init create new users via different mechanisms
- we don't check that the shell & description is sane
- we don't check that the rpm provides are correct

### DIFF
--- a/tests/test_all.py
+++ b/tests/test_all.py
@@ -6,7 +6,9 @@ import datetime
 import fnmatch
 import json
 import pathlib
+import shlex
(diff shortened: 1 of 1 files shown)


alexandrevicenzi:
Many tests fail with the wrong id.

dirkmueller:
was merged via #891

PR #795, opened 2025-04-30
Add openSSL1.1 test in bci-base-fips 15-sp6

pablo-herranz:
Security Certifications team has implicated that the earlier quoted status in ticket [poo#167683](https://progress.opensuse.org/issues/167683) was incorrect, and we do no need also openSSL 1.1 FIPS testing in 15-SP6 as well

[CI:TOXENVS] fips

### DIFF
--- a/tests/test_fips.py
+++ b/tests/test_fips.py
@@ -42,6 +42,11 @@ DOCKERFILE_GCRYPT = """WORKDIR /src/
 COPY tests/files/fips-test-gcrypt.c /src/
 """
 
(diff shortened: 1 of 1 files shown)


dcermak:
Please resubmit from a branch on this repo, not from a fork. Otherwise the secrets will not be setup.

Also, you have created a 1:1 copy of `openssl_fips_hashes_test_fnct`, why? Just use that function in your test case.

And you definitely must add a mark to run this test only on SLE 15 SP6. Currently it will run everywhere

PR #796, opened 2025-05-02
Add 15.7 to the released versions as we are entering GMC

dirkmueller:


alexandrevicenzi:
is this expected?

> No such object: registry.opensuse.org/devel/bci/sle-15-sp7/images/bci/bci-micro:15.7

PR #809, opened 2025-05-06
Add a smoke test checking that readme-url is valid

dcermak:
For https://github.com/SUSE/BCI-dockerfile-generator/pull/2504

### DIFF
--- a/tests/test_metadata.py
+++ b/tests/test_metadata.py
@@ -21,6 +21,7 @@ from typing import Tuple
 
 import pytest
 import requests
+import tenacity
 from _pytest.mark.structures import ParameterSet
 from pytest_container import OciRuntimeBase
 from pytest_container.container import ContainerData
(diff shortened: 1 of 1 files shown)


dirkmueller:
see #430  which you objected to..

dcermak:
> see #430 which you objected to..

So let's not test it?

dirkmueller:
well, the test the way it is written will succeed for opensuse because it fetches a build service generated page afaik that just is containing an error text (but not an error status). which is why there was this convoluted logic in the original PR..

the simplest test to test the PR you looked for is to make sure the url ends in ".md", has "/README" in it and doesn't have "//" inside. no fetching of remote resources needed

dirkmueller:
implemented it the way I suggested here: https://github.com/SUSE/BCI-tests/pull/813

it actually finds  a real problem :sob:

PR #810, opened 2025-05-06
Add docs about the CI & currently required secrets

dcermak:
Required for hand-over. Documents the currently used secrets according to my knowledge, but as @dirkmueller  added them all, I'd need further input.

### DIFF
--- a/source/setup.rst
+++ b/source/setup.rst
@@ -1,27 +1,58 @@
-CI Setup & Secrets/Credentials
+Secrets/Credentials for the CI
 ==============================
 
-The CI is using github actions to run all or a subset of the tests on each pull
-request next to general sanity checks.
+The following outlines the use of Git Hub secrets to enable larger test coverage
+on each pull request as would otherwise be possible. This is because some check
+combinations require credentials to retrieve otherwise access-protected content.
 
-Some checks require access to paywalled LTSS container images on
-registry.suse.com. These are accessible after logging into the registry using
-the credentials ``REGISTRY_LOGIN_USERNAME`` and ``REGISTRY_LOGIN_PASSWORD``.
+Below is a list of secrets and what they are used for.
 
-.. TODO(dirk):
+Access to `registry.suse.com`
+-----------------------------
 
-To regenerate these credentials, go to scc.suse.com, login and navigate to
-https://scc.suse.com/organizations/460227/dashboard. Create a new **XXX**
-subscription starting with ``INTERNAL-USE-ONLY``. The password is the
-registration code and the username is ``regcode``.
+Some checks require access to SUSE Linux Enterprise Server LTSS container images on
+registry.suse.com. These are accessible after logging into the registry which is automatically
+performed by the ci.yml Git Hub Action. It accesses the following secrets stored for that
+purpose:
 
-.. TODO(dirk):
+* ``REGISTRY_LOGIN_USERNAME``
+* ``REGISTRY_LOGIN_PASSWORD``
+
+These point to mirroring credentials in a BCI testing organization. Please contact the
+BCI team for expanding the configuration of that testing organization.
+
+Access to `dp.apps.rancher.io`
+------------------------------
+
+Some checks require access to `SUSE Application Collection <https://apps.rancher.io/>` container images on
+`dp.apps.rancher.io`. These are accessible after logging into the Application Collection Distribution
+Platform ("DP") which is automatically performed by the ci.yml Git Hub Action. It accesses the following
+secrets stored for that purpose:
+
+* ``APPCO_USERNAME``
+* ``APPCO_PASSWORD``
+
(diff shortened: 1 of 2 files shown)


source/setup.rst
```diff
@@ -0,0 +1,27 @@
+CI Setup & Secrets/Credentials
```
dirkmueller, line 1: this is the PR CI? or which CI is it referring to?

source/setup.rst
```diff
@@ -0,0 +1,27 @@
...
+==============================
+
+The CI is using github actions to run all or a subset of the tests on each pull
```
dirkmueller, line 4: NIT GitHub Actions

source/setup.rst
```diff
@@ -0,0 +1,27 @@
...
+request next to general sanity checks.
+
+Some checks require access to paywalled LTSS container images on
```
dirkmueller, line 7:
```suggestion
Some checks require access to access-protected LTSS container images on
```

source/setup.rst
```diff
@@ -0,0 +1,27 @@
...
+registry.suse.com. These are accessible after logging into the registry using
```
dirkmueller, line 8:
```suggestion
`registry.suse.com` or `dp.apps.rancher.io`. These are accessible after logging into the registry using
```

source/setup.rst
```diff
@@ -0,0 +1,27 @@
...
+the credentials ``REGISTRY_LOGIN_USERNAME`` and ``REGISTRY_LOGIN_PASSWORD``.
```
dirkmueller, line 9: There's also APPCO_USERNAME and APPCO_PASSWORD

source/setup.rst
```diff
@@ -0,0 +1,27 @@
...
+
+.. TODO(dirk):
+
+To regenerate these credentials, go to scc.suse.com, login and navigate to
+https://scc.suse.com/organizations/460227/dashboard. Create a new **XXX**
```
dirkmueller, line 14: what kind of org is that? why do we need to log that?

source/setup.rst
```diff
@@ -0,0 +1,27 @@
...
+subscription starting with ``INTERNAL-USE-ONLY``. The password is the
+registration code and the username is ``regcode``.
```
dirkmueller, line 16: no, this won't work. Reason is that I was forced to test all LTSS containers, but LTSS containers have a subscription per arch and version, so you can not do it with a single regcode.

source/setup.rst
```diff
@@ -0,0 +1,27 @@
...
+
+.. TODO(dirk):
+
+The tests for :command:`container-suseconnect` require a
+:file:`/etc/zypp/credentials.d/SCCcredentials` from a registered SLES host. This
```
dirkmueller, line 21: not really, it can also read it from environment variables (and soon directly by using the regcode)

source/setup.rst
```diff
@@ -0,0 +1,27 @@
...
+file is constructed on the CI from the secrets ``SCC_SYSTEM_PASSWORD`` and
+``SCC_SYSTEM_USERNAME``. You can obtain both values by registering a
+``registry.suse.com/bci/bci-base:15.6`` container image via
+:command:`suseconnect --regcode $regcode`. The username and password can then be
```
dirkmueller, line 25: no, you actually need LTSS enabled registration otherwise it'll not work (they are in the public CI only used for the LTSS containers)

source/setup.rst
```diff
@@ -0,0 +1,27 @@
...
+subscription starting with ``INTERNAL-USE-ONLY``. The password is the
+registration code and the username is ``regcode``.
+
+.. TODO(dirk):
+
+The tests for :command:`container-suseconnect` require a
+:file:`/etc/zypp/credentials.d/SCCcredentials` from a registered SLES host. This
+file is constructed on the CI from the secrets ``SCC_SYSTEM_PASSWORD`` and
+``SCC_SYSTEM_USERNAME``. You can obtain both values by registering a
+``registry.suse.com/bci/bci-base:15.6`` container image via
```
dirkmueller, line 24: it's not really recommended to do it that way as the subscription has a short expiry date.

dirkmueller (approved):

dirkmueller:
pushed an update

source/setup.rst
```diff
@@ -0,0 +1,58 @@
+Secrets/Credentials for the CI
+==============================
+
+The following documents the use of Git Hub secrets to enable larger test coverage
```
alexandrevicenzi, line 4:
```suggestion
The following documents outline the use of Git Hub secrets to enable larger test coverage
```

alexandrevicenzi (commented):
organisation (UK) or organization (US)? I prefer the American version.

source/setup.rst
alexandrevicenzi: Yes, because I find the word documents strange in this context.

PR #811, opened 2025-05-06
avoid "in" matches for urls

dirkmueller:
This raises a CodeQL warning. even though it is harmless here as there isn't untrusted input, we can easily avoid it

### DIFF
--- a/tests/test_metadata.py
+++ b/tests/test_metadata.py
@@ -414,13 +414,15 @@ def test_disturl(
     labels = container.inspect.config.labels
 
     disturl = labels["org.openbuildservice.disturl"]
+    baseurl = urllib.parse.urlparse(
+        f"oci://{container.container.get_base().url}"
+    )
     assert (
         disturl
         == labels[
             f"{_get_container_label_prefix(container_name, container_type)}.disturl"
         ]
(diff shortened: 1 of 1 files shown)


tests/test_metadata.py
```diff
@@ -429,7 +428,9 @@ def test_disturl(
             )
         )
     elif OS_VERSION == "16.0":
-        if "opensuse.org" in container.container.get_base().url:
+        if container.container.get_base().url.startswith(
+            "registry.opensuse.org"
```
alexandrevicenzi, line 432: There's no protocol in the URL?

alexandrevicenzi (commented):
The current changes also raise a CodeQL warning.

tests/test_metadata.py
dirkmueller: it's not really a url (dunno why pytest-container calls it that way). it's basically what you would otherwise specify in the `FROM ` statement.

dirkmueller:
> The current changes also raise a CodeQL warning.

*sigh*. well, I had the urlparse before, let me push it again.

alexandrevicenzi (approved):

PR #812, opened 2025-05-06
Add AI tests for released containers

dirkmueller:


alexandrevicenzi (approved):
There are CI failures with bind and openssl, but these seem unrelated to this change, the AI jobs are green.

PR #827, opened 2025-05-20
Add test for bsc#1243360

grisu48:
Add a check if container-suseconnect fails as reaction to https://bugzilla.suse.com/show_bug.cgi?id=1243360

[CI:TOXENVS] all

## Verification runs:

* [SLES 15-SP6 shows the error](https://duck-norris.qe.suse.de/tests/14851#step/_root_BCI-tests_all_/12)
* [SLES 16 without error](https://duck-norris.qe.suse.de/tests/14850#)
* [Leap 15.6 working as well](https://duck-norris.qe.suse.de/tests/14852#)

### DIFF
--- a/tests/test_all.py
+++ b/tests/test_all.py
@@ -19,6 +19,7 @@ from pytest_container import get_extra_build_args
 from pytest_container import get_extra_run_args
 from pytest_container.container import BindMount
 from pytest_container.container import ContainerData
+from pytest_container.runtime import LOCALHOST
 
 from bci_tester.data import ALLOWED_BCI_REPO_OS_VERSIONS
 from bci_tester.data import ALL_CONTAINERS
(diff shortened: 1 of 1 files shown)


tests/test_all.py
```diff
@@ -273,9 +273,16 @@ def test_no_downgrade_on_install(container: ContainerData) -> None:
...
     a question to the user and break the builds.
     """
-
     conn = container.connection
-    conn.run_expect([0], "timeout 2m zypper ref && zypper -n in libsolv-tools")
+
+    zypper_ref = conn.check_output("timeout 2m zypper ref")
+    # Temporary check for bsc#1243360
+    assert (
+        "Skipping service 'container-suseconnect-zypp' because of the above error"
```
dirkmueller, line 281: well, `test_no_downgrade_on_install`is supposed to run without authentication for container-suseconnect ..

grisu48: Check the verification runs, the 15-SP6 catches the container-suseconnect error, the SLES 16 test run is green. This should show that it works as intended or am I missing something?

tests/test_all.py
```diff
@@ -273,9 +274,17 @@ def test_no_downgrade_on_install(container: ContainerData) -> None:
     of any package already installed in the container as that would throw
     a question to the user and break the builds.
     """
-
     conn = container.connection
-    conn.run_expect([0], "timeout 2m zypper ref && zypper -n in libsolv-tools")
+
+    zypper_ref = conn.check_output("timeout 2m zypper ref")
+    # bsc1243360: Check for errors in container-suseconnect-zypper. This check must be limited to SLES hosts
+    if "SLES" in LOCALHOST.check_output("cat /etc/os-release"):
```
alexandrevicenzi, line 281: this fails to catch SL Micro or SLE Micro, but I guess this is catching the QA worker right?

grisu48: Currently we anyways have no test runs on SLEM.

dirkmueller (changes requested):
see inline comment

tests/test_all.py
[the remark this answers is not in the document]
grisu48: But it does, see https://duck-norris.qe.suse.de/tests/14851#step/_root_BCI-tests_all_/12

On registered SLES host, this will always include the additional repositories, irregardless whether they are required on other (non-SLE) hosts or not.

***

Details aside, is your remark that this check for errors with `container-suseconnect-zypper` in `zypper ref` should not be part of `test_no_downgrade_on_install` but it's own test?

grisu48:
Closing this PR. We saw that https://github.com/SUSE/BCI-tests/pull/826 works so there is no need for this anymore.

PR #829, opened 2025-05-21
Limit subscription repo tests to SLES hosts

grisu48:
The check for additional repositories should only run on SLES hosts, where subscriptions are available.

Verification runs: [Leap (skipped)](https://duck-norris.qe.suse.de/tests/14859#step/_root_BCI-tests_all_/17) | [SLES](https://duck-norris.qe.suse.de/tests/14860#step/_root_BCI-tests_all_/17)

[CI:all]

### DIFF
--- a/tests/test_all.py
+++ b/tests/test_all.py
@@ -7,6 +7,7 @@ import fnmatch
 import json
 import pathlib
 import xml.etree.ElementTree as ET
+from pathlib import Path
 
 import packaging.version
 import pytest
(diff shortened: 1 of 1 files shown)


tests/test_all.py
```diff
@@ -742,6 +743,9 @@ def test_container_suseconnect_adds_repos(container_per_test: ContainerData):
     (SLE_BCI, SLE_BCI_debug and SLE_BCI_source).
 
     """
+    # Skip test on non-SLES hosts
+    if "SLES" not in LOCALHOST.check_output("cat /etc/os-release"):
+        pytest.skip("subscription tests must only run on SLES hosts")
```
dcermak, line 748: That's not what you want, what you actually want is to correct line 731 to:
```
not Path(ZYPP_CREDENTIALS_DIR).exists() or len(list(Path(ZYPP_CREDENTIALS_DIR).iterdir())) == 0
```
Yes, that's my fault. Then Leap should be skipped too, as long as it is unregistered.

My mistake was to only skip this test if /etc/zypp/credentials.d is absent, not when it is present, but empty

grisu48: I think that relying on the zypper credentials dir is not what we want because it makes the test prone to false negatives. If for some reason `/etc/zypp/credentials.d` is not present or created, the test would be wrongly skipped.

It's IMHO safer to explicitly state that this test MUST run on SLES and not rely on indirect assumptions that on SLES that directory must be present.

grisu48:
VRs restarted.

dcermak (approved):

alexandrevicenzi (approved):

PR #830, opened 2025-05-21
Switch bind test to opensuse.org

dcermak:
The domain is more resilient to outages. However, opensuse.org has multiple MX records and therefore we cannot compare just the first record from the bind container to the first from the host, we must compare them all.

### DIFF
--- a/tests/test_bind.py
+++ b/tests/test_bind.py
@@ -3,7 +3,7 @@
 from datetime import timedelta
 from pathlib import Path
 from typing import Dict
-from typing import Optional
+from typing import List
 from typing import Union
 
 import dns.query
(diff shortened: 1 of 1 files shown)


tests/test_bind.py
```diff
@@ -49,9 +49,10 @@ def test_basic_resolution(
     for the ``suse.com`` domain for the ``A``, ``AAAA`` and ``MX`` records.
```
alexandrevicenzi, line 49:
```suggestion
    for the ``opensuse.com`` domain for the ``A``, ``AAAA`` and ``MX`` records.
```

alexandrevicenzi (approved):

grisu48 (approved):

PR #834, opened 2025-05-26
Inject SCC subscription via environment variables

dirkmueller:
This allows it to be compatible with SLES16 and avoids having to run tests as root.


grisu48 (approved):
LGTM!

PR #845, opened 2025-06-03
Add testing for kubernetes 1.33

dirkmueller:


dirkmueller:
> I see this error: `'kubectl_1.33' not found in `markers` configuration option`

fixed

PR #852, opened 2025-06-17
Limit legacy openssl tests to SLES hosts

grisu48:
FIPS-based legacy openssl tests require the Legacy Module and thus a SLES subscription. So we need to limit those tests to SLES hosts only.

* Related failures: https://openqa.suse.de/tests/18082593#step/_root_BCI-tests_base_/6
* Verification runs: https://duck-norris.qe.suse.de/tests/14874

[CI:TOXENVS] fips


tests/test_fips.py
```diff
@@ -179,8 +180,12 @@ def run_digest_tests(command_prefix: str):
 
     run_digest_tests("openssl")
 
-    # SLE 15-SP6 has to be tested with OpenSSL v1.1 and v3
-    if OS_VERSION == "15.6":
+    # SLE 15-SP6 has to be tested with OpenSSL v1.1 and v3.
+    # Test module required Legacy Modules so this can only run on SLES hosts
+    if (
+        "SUSE Linux Enterprise Server" in OS_PRETTY_NAME
```
dirkmueller, line 186: OS_PRETTY_NAME is defining the *expected* os of the *container, not of the host. 

why don't you simply check whether container-suseconnect has valid credentials? of better, inject proper credentials also on on-sles hosts? (either via the environment variables or by adding a /etc/zypp/credentials.d/SCCcredentials file on the host)

dirkmueller (changes requested):
see inline

tests/test_fips.py
grisu48: Changed it accordingly.

At the moment only SLES hosts hold credentials and there are no plans to change this any time soon. So we can rely on checking if the host system is SLES or not.

dirkmueller: ok, better. but we're still hardcoding the assumption of the QA setup into the testsuite, while I think we could just test whether things are available and otherwise just skip the test. any problem with that?

PR #859, opened 2025-06-25
switch to pypi releases for pytest_container

dirkmueller:

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -5,6 +5,7 @@ skip_missing_interpreters = True
 [common]
 deps =
     pytest
+    pytest-container >= 0.4.3
     pytest-testinfra
     pytest-xdist ; python_version >= "3.6"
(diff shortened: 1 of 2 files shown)


tox.ini
```diff
@@ -5,6 +5,7 @@ skip_missing_interpreters = True
 [common]
 deps =
     pytest
+    pytest-container >= 0.4.3
```
alexandrevicenzi, line 8: are we sure we want >= here? it might pick a version that breaks compatibility

dirkmueller: > are we sure we want >= here? it might pick a version that breaks compatibility

when we notice that, we can introduce a `<` or `!=` condition (or decide to swallow the incompatibility and upgrade

alexandrevicenzi (approved):

PR #861, opened 2025-06-30
Add xorg-client container

dirkmueller:


alexandrevicenzi (changes requested):
There are some metadata failures for xorg client on SP7.

PR #863, opened 2025-07-01
Add check for trademark being in place (jsc#PED-12111)

dirkmueller:


dirkmueller:
(waiting until SLE16 part is done)

PR #867, opened 2025-07-01
Enable BCI-repo tests for SL16

dirkmueller:


alexandrevicenzi (commented):
Looks good, but there seems to be an issue with repos, causing package choices.

ricardobranco777 (approved):

PR #871, opened 2025-07-07
Check for leftover files in the container

dirkmueller:


alexandrevicenzi (commented):
Overall looks good.

Can you rebase so we can merge it?

PR #884, opened 2025-08-15
Fix EULA assertion

volodymyrkatkalov:
Related ticket: https://progress.opensuse.org/issues/187371
Verification run:  https://openqa.suse.de/tests/18792907


dirkmueller:
no, we want to only check for the BETA eula on beta distros and only the release eula on release distros

PR #891, opened 2025-08-26
Add a smoke test that all users in the container have a stable uid & gid

rcmadhankumar:
First draft to test whether uids & gids are stable. Currently this should pass and establishes a baseline that we can rely on, but has some limitations:

    the entrypoints of pcp & init create new users via different mechanisms
    we don't check that the shell & description is sane
    we don't check that the rpm provides are correct

Co-authored-by: dcermak <dcermak@suse.com>


tests/test_all.py
```diff
@@ -763,3 +768,214 @@ def test_container_suseconnect_adds_repos(container_per_test: ContainerData):
...
+        assert uid == expected_uid, (
+            f"Expected user {name} to have uid {expected_uid} but got {uid}"
+        )
+        assert gid == expected_gid, (
+            f"Expected user {name} to have gid {expected_gid} but got {gid}"
+        )
+
+
+@pytest.mark.parametrize("container", ALL_CONTAINERS, indirect=True)
+def test_all_users_provided_by_sysusers(container: ContainerData) -> None:
```
alexandrevicenzi, line 892: This test is failing for a few images

tests/test_all.py
```diff
@@ -763,3 +768,214 @@ def test_container_suseconnect_adds_repos(container_per_test: ContainerData):
...
+    _USERNAME_UID_GID_MAP["pcp"] = (496, 498)
+    _USERNAME_UID_GID_MAP["pcpqa"] = (494, 498)
+    _USERNAME_UID_GID_MAP["messagebus"] = (498, 498)
+    _USERNAME_UID_GID_MAP["wwwrun"] = (None, 498)
+    _USERNAME_UID_GID_MAP["pesign"] = (499, 499)
+    _USERNAME_UID_GID_MAP["systemd-coredump"] = (497, 1000)
+
+
+@pytest.mark.parametrize("container", ALL_CONTAINERS, indirect=True)
+def test_uids_stable(container: ContainerData) -> None:
```
alexandrevicenzi, line 827: This test is failing

> Expected user messagebus to have uid 498 but got 499

alexandrevicenzi (commented):
The CI is running for all envs, and it fails for many.

Add `[CI:TOXENVS] minimal` to the PR description to run only a subset of tests, e.g skip FIPS.

tests/test_all.py
```diff
@@ -763,3 +768,240 @@ def test_container_suseconnect_adds_repos(container_per_test: ContainerData):
...
+            "opensuse/samba-server:latest"
+        ) and name == "messagebus":
+            expected_gid = 498
+            expected_uid = 498
+
+        if (container.container.get_base().baseurl or "").endswith(
+            "opensuse/samba-toolbox:latest"
+        ) and name == "messagebus":
+            expected_gid = 498
+            expected_uid = 498
```
dirkmueller, line 878: this seems to be duplicated code, can you combine them into one if condition?

tests/test_all.py
```diff
@@ -763,3 +768,240 @@ def test_container_suseconnect_adds_repos(container_per_test: ContainerData):
...
+        if (container.container.get_base().baseurl or "").endswith(
+            "opensuse/samba-client:latest"
+        ) and name == "messagebus":
+            expected_gid = 499
+            expected_uid = 499
+
+        if (container.container.get_base().baseurl or "").endswith(
+            "opensuse/samba-server:latest"
+        ) and name == "messagebus":
+            expected_gid = 498
```
dirkmueller, line 871:
```suggestion
            expected_uid, expected_gid = (498, 498)
```

is imho a bit better to read.

tests/test_all.py
```diff
@@ -763,3 +768,203 @@ def test_container_suseconnect_adds_repos(container_per_test: ContainerData):
...
+    _USERNAME_UID_GID_MAP["systemd-coredump"] = (497, 1000)
+
+
+@pytest.mark.parametrize("container", ALL_CONTAINERS, indirect=True)
+def test_uids_stable(container: ContainerData) -> None:
+    """Check that every user in :file:`/etc/passwd` has a stable uid & gid as
+    defined in ``_USERNAME_UID_GID_MAP``.
+
+    """
+    # collect users who owns subdirectories in directories /var, /etc, /opt, home
```
alexandrevicenzi, line 817:
```suggestion
    # collect users who owns subdirectories in directories /var, /etc, /opt, /home
```

tests/test_all.py
```diff
@@ -763,3 +768,240 @@ def test_container_suseconnect_adds_repos(container_per_test: ContainerData):
...
+        )
+        assert gid == expected_gid, (
+            f"Expected user {name} to have gid {expected_gid} but got {gid}"
+        )
+
+
+@pytest.mark.parametrize("container", ALL_CONTAINERS, indirect=True)
+def test_all_users_provided_by_sysusers(container: ContainerData) -> None:
+    """Check that all users from :file:`/etc/passwd` are provided by a
+    corresponding conf file in :file:`/usr/lib/sysusers.d/`.
```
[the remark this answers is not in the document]
alexandrevicenzi: But this is the point of switching to a user in certain containers, no? We would create users with useradd, and not systemd.

There's already a special case `users_not_via_sysusers_d` and this list would possibly grow, causing this test to be less relevant perhaps.

tests/test_all.py
```diff
@@ -763,3 +768,203 @@ def test_container_suseconnect_adds_repos(container_per_test: ContainerData):
...
+        ].endswith("kiosk/xorg") and name == "user":
+            expected_gid = 100
+
+        if (
+            (container.container.get_base().baseurl or "")
+            .split(":")[0]
+            .endswith("opensuse/pcp")
+        ):
+            if name == "pcp":
+                expected_gid, expected_uid = (498, 496)
```
alexandrevicenzi, line 852: Isn't this the same as `_USERNAME_UID_GID_MAP["pcp"] = (496, 498)`?

tests/test_all.py
```diff
@@ -763,3 +768,203 @@ def test_container_suseconnect_adds_repos(container_per_test: ContainerData):
...
+        expected_uid = 499 if expected_uid is None else expected_uid
+        expected_gid = 499 if expected_gid is None else expected_gid
+
+        # special cases
+        if (container.container.get_base().baseurl or "").split(":")[
+            0
+        ].endswith("kiosk/xorg") and name == "user":
+            expected_gid = 100
+
+        if (
```
alexandrevicenzi, line 846: Can we move this to a fixture? There's `_USERNAME_UID_GID_MAP` for generic users, maybe `_USERNAME_UID_GID_IMAGE_MAP` could map image names vs users for special cases.

It would remove these IFs here and be in one single place. To update in the future, it would be easier.

tests/test_all.py
rcmadhankumar: yes, fixed it.

dirkmueller (approved):

PR #895, opened 2025-09-01
URGENT: Use pytest-rerunfailures < 16.0

grisu48:
pytest-rerunfailures 16.0 breaks all test runs that use it, so we need to stay on the previous version for now.

* Ticket reference: https://progress.opensuse.org/issues/188013


grisu48:
I tested this in a local Tumbleweed VM and could resolve the issue with this update.

ricardobranco777 (approved):

PR #910, opened 2025-09-26
Switch to openjdk25 on tumbleweed

dirkmueller:


dirkmueller:
> This test change seems to cause failure on tumbleweed: https://openqa.opensuse.org/tests/5346799#step/bci_test_podman/43

the failure is:

```subprocess.CalledProcessError: Command '['podman', 'pull', 'registry.opensuse.org/opensuse/factory/totest/containers/opensuse/apache-tomcat:10.1-openjdk25']'
``` 

which means the respective container isn't there yet.

PR #914, opened 2025-09-30
Add RMT server for testing

dirkmueller:
[CI:TOXENVS] all


grisu48 (approved):
LGTM

I tested both the `all` and `metadata` environment, both work for the `rmt` container image.

PR #1035, opened 2026-03-31
Skip downgrade-tests on SLES 16.1 unconditionally

grisu48:
There are no SLES 16.1 repos available until public beta, so the downgrade tests needs to be skipped until then.

Related failure: https://openqa.suse.de/tests/21598666#step/_root_BCI-tests_all_/14

### DIFF
--- a/tests/test_all.py
+++ b/tests/test_all.py
@@ -51,7 +51,6 @@ from bci_tester.data import OS_VERSION_ID
 from bci_tester.data import PCP_CONTAINERS
 from bci_tester.data import RELEASED_LTSS_VERSIONS
 from bci_tester.data import RELEASED_SLE_VERSIONS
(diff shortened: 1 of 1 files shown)


dirkmueller:
> > weird, this isn't supposed to happen. let me investigate. in the worst case I would prefer to include the x86-64-v3 package as well. I'd argue that containers must not break the minimal system requirements as well.
> That cannot be done, as SLES16 only has [x86_64 v2](https://documentation.suse.com/releasenotes/sles/html/releasenotes_sles_16.0/index.html) as minimal requirement. I'd argue that the SLES 16.1 containers must also follow the minimal system requirements. The v3 libraries are only optional.

sure it can be done. the v3 version of the libraries are only used on a v3 system, otherwise it will use the v2 libraries.

PR #1043, opened 2026-04-20
default group of added userfs changed from 100 to 1000 on sp7

msmeissn:


dirkmueller:
that's a potentially breaking change? can we revert the gid change in the OS?

dirkmueller:
hardcoded gid 100 in the container for now.

dirkmueller:
sorry, this was another user for coredump

dirkmueller:
moved the partial to #1044

PR #1044, opened 2026-04-21
default group of added userfs changed from 100 to 1000 on sp7

dirkmueller:


alexandrevicenzi (dismissed):

tests/test_all.py
alexandrevicenzi: nit: we can remove from here

alexandrevicenzi (approved):

PR #1046, opened 2026-04-24
Fix no tests ran for python-micro

alexandrevicenzi:
See bsc#1262667.

[CI:TOXENVS] all,metadata,minimal,python


dirkmueller (changes requested):
It should be added to containers without zypper then

alexandrevicenzi:
> It should be added to containers without zypper then

where? it was added to it

dirkmueller (approved):

dirkmueller:
> > It should be added to containers without zypper then
> where? it was added to it

sorry, got confused (reviewed this on the phone). it's correct.

PR #1057, opened 2026-05-05
Skip downgrade tests while repo is broken (bsc#1263990)

dirkmueller:
[CI:TOXENVS] all


alexandrevicenzi (approved):

alexandrevicenzi:
Maybe we need a similar skip on `test_no_orphaned_packages`, because it has conflicts also.

dirkmueller:
I think thats  a bug worth fixing as well though..

PR #1061, opened 2026-05-07
Move buildx check after Docker check

alexandrevicenzi:
If the runtime is not Docker, docker might not be installed.

[CI:TOXENVS] go,minimal


dirkmueller (changes requested):
not working, as the condition of skipif() itself is evaluated regardless..

PR #1069, opened 2026-05-19
Fix tests for 16.1

dirkmueller:
[CI:TOXENVS] go,mariadb


alexandrevicenzi:
Something broke, golang test is failing on everything else.

dirkmueller:
> Something broke, golang test is failing on everything else.

because the target is called "go"

alexandrevicenzi (approved):

PR #1072, opened 2026-05-27
Increase Base size for 16.1

alexandrevicenzi:
[CI:TOXENVS] base,minimal

### DIFF
--- a/tests/test_base.py
+++ b/tests/test_base.py
@@ -176,7 +176,14 @@ def test_base_size(container: ContainerData, container_runtime):
             "ppc64le": 130,
             "s390x": 95,
         }
-    elif OS_VERSION in ("16.0", "16.1"):
+    elif OS_VERSION in ("16.1",):
+        base_container_max_size: Dict[str, int] = {
+            "x86_64": 96,
+            "aarch64": 101,
+            "ppc64le": 117,
+            "s390x": 93,
+        }
+    elif OS_VERSION in ("16.0",):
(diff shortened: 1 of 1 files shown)


alexandrevicenzi:
Caught by:

```
FAILED tests/test_base.py::test_base_size[bci/bci-base:16.1 from registry.opensuse.org/devel/bci/16.1/containerkiwi/bci/bci-base:16.1] - AssertionError: Base container size is 96.0 MiB for x86_64 (expected max of 95 MiB)
```

tests/test_base.py
```diff
@@ -176,7 +176,14 @@ def test_base_size(container: ContainerData, container_runtime):
             "ppc64le": 130,
             "s390x": 95,
         }
-    elif OS_VERSION in ("16.0", "16.1"):
+    elif OS_VERSION in ("16.1",):
+        base_container_max_size: Dict[str, int] = {
+            "x86_64": 96,
+            "aarch64": 100,
```
dirkmueller: probably also have to update this by one MB. did you btw check whether the filelist is the same and where the increase is coming from?

alexandrevicenzi: compare to what exactly? there are no older builds (history) for registry.opensuse.org/devel/bci/16.1/containerkiwi/bci/bci-base so it makes a bit difficult to find the real reason, I tried to compare to other images, but there were too many diffs.

dirkmueller (approved):

PR #1073, opened 2026-05-27
Skip NSS test on Tumbleweed (boo#1266391)

alexandrevicenzi:
[CI:TOXENVS] fips,minimal

### DIFF
--- a/tests/test_fips.py
+++ b/tests/test_fips.py
@@ -397,6 +397,10 @@ def test_icainfo_binary(container_per_test: ContainerData) -> None:
     container_per_test.connection.check_output("icastats -S")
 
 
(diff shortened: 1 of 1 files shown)


tests/test_fips.py
```diff
@@ -397,6 +397,10 @@ def test_icainfo_binary(container_per_test: ContainerData) -> None:
     container_per_test.connection.check_output("icastats -S")
 
 
+@pytest.mark.skipif(
```
dirkmueller: please use `pytest.mark.xfail` so that it fails once the problem is fixed.

dirkmueller (changes requested):

dirkmueller (approved):

PR #1076, opened 2026-06-10
Add check to ensure we have no transient logs in the container

dirkmueller:
[CI:TOXENVS] all

### DIFF
--- a/tests/test_all.py
+++ b/tests/test_all.py
@@ -940,3 +940,19 @@ def test_uids_stable(auto_container) -> None:
         assert gid == expected_gid, (
             f"Expected user {name} to have gid {expected_gid} but got {gid}"
         )
+
+
+def test_no_transient_logfiles(auto_container):
+    """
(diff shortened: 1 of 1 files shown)


tests/test_all.py
```diff
@@ -926,3 +930,15 @@ def uid_gid_map(container: ContainerData) -> Dict[str, Tuple[int, int]]:
...
             f"Expected user {name} to have gid {expected_gid} but got {gid}"
         )
+
+
+def test_no_transient_logfiles(auto_container):
+    """
+    Ensure that no transient state files are leftover in the container from
+    the build process.
+    """
+    assert not auto_container.connection.file(
```
alexandrevicenzi: nit: this could be a for loop over a list of files, no?

alexandrevicenzi (dismissed):

tests/test_all.py
dirkmueller: sure, done

alexandrevicenzi (approved):

PR #1077, opened 2026-06-12
Update Go TLS test to be more resilient

alexandrevicenzi:
Instead of trying only suse.com, try at least 5 domains, and if 3 work out, consider it a success.

[CI:TOXENVS] all,minimal

### DIFF
--- a/tests/test_all.py
+++ b/tests/test_all.py
@@ -60,17 +60,49 @@ from bci_tester.util import is_spr
 CONTAINER_IMAGES = [x for x in ALL_CONTAINERS if x != NANO_CONTAINER]
 
 
-#: Go file to perform a GET request to suse.com and that panics if the request
-#: fails
-FETCH_SUSE_DOT_COM = """package main
+# Go file to perform GET requests and check if 3 out of 5 requests succeed.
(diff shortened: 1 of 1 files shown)


tests/test_all.py
```diff
@@ -60,17 +60,49 @@
...
+    "os"
+    "time"
+)
 
 func main() {
-        _, err := http.Get("https://suse.com/")
-        if err != nil {
-                panic(err)
+        urls := []string{
+            "https://suse.com",
```
dirkmueller, line 76: do we actually have an issue with suse.com  rather than a client side thing?

I would recommend we use something that we know is available. you could use for example `https://updates.suse.com/-/healthy` .

tests/test_all.py
```diff
@@ -60,17 +60,49 @@
...
+            "https://google.com",
+            "https://cloudflare.com",
+            "https://github.com",
+            "https://letsencrypt.org",
```
dirkmueller, line 80: no, we are not spamming a non-profit third party NGO with our testsuites.

dirkmueller (changes requested):

PR #1079, opened 2026-06-15
test_fips: handle zypper ref failures from appdata download timeouts

dirkmueller:
The `tox (fips, tumbleweed)` CI job was failing because `zypper ref` exited with code 4 (`ZYPPER_EXIT_ERR_ZYPP`) when downloading repository metadata inside the `bci-base-fips:latest` (Tumbleweed) container.

## Root Cause

The `openSUSE-Tumbleweed-Oss` repository's `appdata.xml.gz` (AppStream metadata, ~200–500 MB) timed out downloading from `cdn.opensuse.org` (Curl error 28). When any metadata file listed in `repomd.xml` fails to download, zypper marks the entire repository as invalid and exits with code 4, even though the package metadata (`primary.xml.gz`) needed for installation had already been successfully cached.

## Fix

In `tests/test_fips.py`, all `zypper ref` calls are changed from:

```bash
zypper --gpg-auto-import-keys -n ref && zypper -n install ...
```

to:

```bash
zypper --gpg-auto-import-keys -n ref; zypper -n ref -B && zypper -n install ...
```

The semicolon ensures `zypper ref -B` (build-only database rebuild) always runs after the initial refresh. When the initial ref fails partway through (e.g., after downloading `primary.xml.gz` but before the large `appdata.xml.gz` completes), `zypper ref -B` rebuilds the package database from the successfully-cached metadata, making the repository usable for subsequent `zypper install` operations.

This fix applies to `test_openssl_binary`, `test_gnutls_binary`, `test_gcrypt_binary`, and `test_nss_firefox_cert`.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -129,6 +129,7 @@ jobs:
         os_version:
           - 15.7
           - "16.0"
+          - "16.1"
           - "tumbleweed"
         include:
           - toxenv: fips


alexandrevicenzi (dismissed):

dirkmueller:
Thanks, AI.

PR #1091, opened 2026-07-08
Prepare handling for kubevirt 1.8

dirkmueller:
[CI:TOXENVS] all


alexandrevicenzi:
it appears that it fails to pull the images on 16.1? not sure if the image does not exist or the registry is having issues

dirkmueller:
> it appears that it fails to pull the images on 16.1? not sure if the image does not exist or the registry is having issues

kubevirt 1.8 is still in process of being added to 16.1.

alexandrevicenzi (approved):

PR #1110, opened 2026-07-29
Test the packaged tox on slfo as well

dirkmueller:
[CI:TOXENVS]  python

### DIFF
--- a/tests/test_python.py
+++ b/tests/test_python.py
@@ -197,9 +197,6 @@ def test_tox(auto_container_per_test):
     )
 
 
-@pytest.mark.skipif(
-    OS_VERSION in ["16.0", "16.1"], reason="no packaged tox version available"
-)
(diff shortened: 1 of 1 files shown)


alexandrevicenzi:
Something is off, is the package/binary name correct?

It fails with `/bin/sh: line 1: python3: command not found` and `Package 'python3-tox' not found.`

dirkmueller:
> It fails with `/bin/sh: line 1: python3: command not found` and `Package 'python3-tox' not found.`

that's just the containers now being generally broken thanks to the python3/python313 rework in SL16.0

msmeissn:
either use python3.13   as interpreter, or install also python3 additionaly.

dirkmueller:
> either use python3.13 as interpreter, or install also python3 additionaly.

yes, already fixed

PR #1119, opened 2026-08-12
Fix tox setup

alexandrevicenzi:
Avoid using sudo, as it overrides system packages.


grisu48 (approved):

alexandrevicenzi:
There are many CI failures, timeouts, or failures to pull images, but they do not look related to my changes.

PR #1122, opened 2026-08-13
ruby: force install rake on 16.1

alexandrevicenzi:
16.1 uses alts, and the installation refuses to override system binaries.

[CI:TOXENVS] minimal, ruby


mloviska (approved):

alexandrevicenzi:
#1092 drops the test, which is better than my proposal.

PR #1123, opened 2026-08-18
fips: ensure Micro FIPS is configured properly (bsc#1274584)

alexandrevicenzi:
If FIPS is not configured properly, non-FIPS digests will succeed and fail the test run.

[CI:TOXENVS] minimal, base, fips


grisu48 (approved):
The Multi-Build idea is really nice.

dirkmueller (approved):

PR #1129, opened 2026-08-24
fips: avoid strict output checking

alexandrevicenzi:
Test only if it succeeds or fails, without strictly checking blank outputs. Otherwise, warnings might break the tests.

[CI:TOXENVS] minimal, fips


alexandrevicenzi:
Caught on https://openqa.suse.de/tests/24036466#step/bci_test_podman/1

mloviska (approved):

grisu48 (approved):

dirkmueller:
> Caught on https://openqa.suse.de/tests/24036466#step/bci_test_podman/1

this warning is also visible to users, no? we should maybe not skip over it..

PR #1132, opened 2026-08-26
ruby: Add libexslt0 to SLE 16 (bsc#1276559)

alexandrevicenzi:
nokogiri requires libexslt0 and rails template fails without it.

[CI:TOXENVS] minimal, ruby

### DIFF
--- a/tests/test_ruby.py
+++ b/tests/test_ruby.py
@@ -93,17 +93,10 @@ def test_rails_template(auto_container_per_test):
     ).startswith("2.5"):
         pytest.xfail("Rails 8 needs Ruby >= 3.0")
 
-    # Rails asset pipeline needs Node.js and yarn. apps:template needs libxslt
-    auto_container_per_test.connection.check_output(
-        "zypper -n in nodejs-default libyaml-devel libxslt1"
-    )
-
     auto_container_per_test.connection.check_output(
         "gem install 'rails:~> 8.0'"
     )
 
-    # auto_container_per_test.connection.run_expect([0], "zypper -n in npm nodejs")
-    # auto_container_per_test.connection.run_expect([0], "npm -g install yarn")
     auto_container_per_test.connection.check_output(
         "rails new /hello/ --minimal"
     )


dirkmueller:
we should do this in the container itself, not just in the tests imho. the container should be suitable for rails deployment.

Vogtinator:
I think this is actually a nokogiri bug, same as https://github.com/sparklemotion/nokogiri/issues/1562 in the past

Vogtinator:
It's actually a ruby packaging issue: `--disable-rpath` breaks building nokogiri with its bundled libs from source. It also fails on 15.7 in the current `registry.suse.com/bci/ruby:3.4` container, I wonder why that wasn't found earlier.

alexandrevicenzi:
> we should do this in the container itself, not just in the tests imho. the container should be suitable for rails deployment.

It seems that libxslt1 is also a dependency, so I just followed the same principle. True, Rails should work out of the box, but unclear to me what is required for Rails to work since I have never used it.

alexandrevicenzi:
> It's actually a ruby packaging issue: `--disable-rpath` breaks building nokogiri with its bundled libs from source. It also fails on 15.7 in the current `registry.suse.com/bci/ruby:3.4` container, I wonder why that wasn't found earlier.

I did check 15.7, but the package libexslt0 does not even exist there.

alexandrevicenzi:
> I think this is actually a nokogiri bug, same as [sparklemotion/nokogiri#1562](https://github.com/sparklemotion/nokogiri/issues/1562) in the past

Interesting, I would need to look deeper into it and see if it applies here.

Vogtinator:
> > I think this is actually a nokogiri bug, same as [sparklemotion/nokogiri#1562](https://github.com/sparklemotion/nokogiri/issues/1562) in the past
> 
> Interesting, I would need to look deeper into it and see if it applies here.

See the next comment:

> It's actually a ruby packaging issue: `--disable-rpath` breaks building nokogiri with its bundled libs from source. It also fails on 15.7 in the current `registry.suse.com/bci/ruby:3.4` container, I wonder why that wasn't found earlier.

PR #1138, opened 2026-09-02
fips: add version check

alexandrevicenzi:
Check if both bci-base-fips and bci-micro-fips have the same libopenssl3 version.

[CI:TOXENVS] minimal, fips

### DIFF
--- a/tests/test_fips.py
+++ b/tests/test_fips.py
@@ -15,9 +15,11 @@ from pytest_container import get_extra_build_args
 from pytest_container import get_extra_run_args
 from pytest_container.container import BindMount
 from pytest_container.container import ContainerData
+from pytest_container.container import ContainerLauncher
 from pytest_container.container import ImageFormat
 from pytest_container.container import container_and_marks_from_pytest_param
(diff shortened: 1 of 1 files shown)


tests/test_fips.py
```diff
@@ -522,3 +524,39 @@ def test_fips_properly_setup_on_micro(
...
+        base_con = base_launcher.container_data.connection
+
+        micro_ver = micro_con.check_output(
+            'file=$(ls /usr/lib64/libssl.so.3.*); echo "${file##*libssl.so.}"'
+        ).strip()
+        base_ver = base_con.check_output(
+            'file=$(ls /usr/lib64/libssl.so.3.*); echo "${file##*libssl.so.}"'
+        ).strip()
+
+        assert micro_ver == base_ver, (
```
dirkmueller, line 562: We should check that it is non empty and not have an expansion error. Because then the string is likely the same

alexandrevicenzi: Good catch. I added a simple length check.

dirkmueller (approved):

PR #1142, opened 2026-09-17
ruby: add micro variant

alexandrevicenzi:
[CI:TOXENVS] minimal, ruby


dirkmueller (approved):

DimStar77:
openQA was apparently unaware of that - and still tries to test ruby_latest, resulting in failures
https://openqa.opensuse.org/tests/6226899

PR #1144, opened 2026-09-17
fix: golang stable needs to be the last

alexandrevicenzi:
Rework the Go definition to ensure oldstable and stable always appear last in the list.

[CI:TOXENVS] minimal, multistage, go


dirkmueller (approved):
thanks

